PR #119, opened 2025-05-06
Update SQLAlchemy example to include non-admin and connection refresh logic

mitchell-elholm:
*Description of changes:*

Adds non-admin logic and connection token refresh
- Adds a listener for non-admin, and updates the search path to myschema.
- Adds a listener to set a new token every time a connection is created. The connection is created immediately after the parameter is updated.
- Connection refresh can be tested by setting the token expiry to 2 seconds, and then setting a parameter pool_recycle to 1 second in the engine creation. This ensures that it will attempt to open a new connection every second, and that the first token created will not be valid anymore after 1 second. It should succeed, as it created a new token each time it connected. I also tested the negative case, by removing the listener altogether and using a single token with a short timeout, to prove that it would have failed without the token refresh.
```
engine = create_engine(
        url, 
        connect_args={"sslmode": "verify-full", "sslrootcert": "system"},
        pool_recycle=1 
    )
...

return client.generate_db_connect_admin_auth_token(cluster_endpoint, region, 1)
```
- Other minor changes: Adds table drop to allow repeated execution.
- Removed SDK, added import

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/python/sqlalchemy/README.md
+++ b/python/sqlalchemy/README.md
@@ -1,30 +1,285 @@
-# Aurora DSQL SQLAlchemy code examples
+# SQLAlchemy with Aurora DSQL
 
 ## Overview
 
-The code examples in this topic show you how to use the SQLAlchemy work with Aurora DSQL. 
+This code example demonstrates how to use SQLAlchemy with Amazon Aurora DSQL. The example shows you how to
+connect to an Aurora DSQL cluster with SQLAlchemy using Psycopg2, create entities, and read and write to those entity tables.
 
-## Run the examples
+Aurora DSQL is a distributed SQL database service that provides high availability and scalability for
+your PostgreSQL-compatible applications. SQLAlchemy is a popular object-relational mapping framework for Python that allows
+you to persist Python objects to a database while abstracting the database interactions.
+
+> **Note**
+>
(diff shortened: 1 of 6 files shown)


python/sqlalchemy/README.md
```diff
@@ -1,30 +1,285 @@
...
+connect to an Aurora DSQL cluster with SQLAlchemy using Psycopg2, create entities, and read and write to those entity tables.
 
-## Run the examples
+Aurora DSQL is a distributed SQL database service that provides high availability and scalability for
+your PostgreSQL-compatible applications. SQLAlchemy is a popular object-relational mapping framework for Python that allows
+you to persist Python objects to a database while abstracting the database interactions.
+
+> **Note**
+>
+> Note that SQLAlchemy with Psycopg3 does not work with Aurora DSQL. SQLAlchemy with Psycopg3 uses nested transactions which rely on savepoints as part of the connection setup. Savepoints are not supported by Aurora DSQL.
```
trstephen-amazon, line 14: TIL

trstephen-amazon (approved):
🚢

PR #123, opened 2025-05-09
Update cpp libpq samples and README.md

leszek-bq:
This PR updates the cpp libpq example

Description of changes:

- Add non-admin connection handing
- Update the documentation in README.md

### DIFF
--- a/.github/workflows/cpp-libpq-integ-tests.yml
+++ b/.github/workflows/cpp-libpq-integ-tests.yml
@@ -77,6 +77,7 @@ jobs:
       env:
         CLUSTER_ENDPOINT: ${{ secrets.CPP_LIBPQ_CLUSTER_ENDPOINT }}
         REGION: ${{ secrets.CPP_LIBPQ_CLUSTER_REGION }}
+        CLUSTER_USER: admin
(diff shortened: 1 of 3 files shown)


.github/workflows/cpp-libpq-integ-tests.yml
```diff
@@ -77,6 +77,7 @@ jobs:
       env:
         CLUSTER_ENDPOINT: ${{ secrets.CPP_LIBPQ_CLUSTER_ENDPOINT }}
         REGION: ${{ secrets.CPP_LIBPQ_CLUSTER_REGION }}
+        CLUSTER_USER: admin
```
imforster, line 80: Can we also do a second pass that runs with non-admin user also?

leszek-bq: We currently don't do that for other languages. This could be next week addition.

imforster: Let's create Taskei to ensure we don't forget this.

leszek-bq: We'll create a taskei ticket to track this for all languages.

imforster (approved):

PR #128, opened 2025-05-13
java: refactor SR cluster management

trstephen-amazon:
#### Changes
- Deletes MR examples; this will be added back once SDKs are available with updated APIs
- Combines all SR actions into a single test case. This'll let CI run faster because we don't have to create a cluster for each operation.
- Adds a sweeper that deletes leaked clusters. Only runs in CI.
- Remove unused and duplicate dependencies.
- Readme corresponds to this code instead of pgJDBC ;)

----


trstephen-amazon:
bah, mis-clicked the 'ready for review' button. It's _probably_ ready but I want to see CI go green.

java/cluster_management/src/main/java/org/example/CreateCluster.java
```diff
@@ -1,53 +1,54 @@
...
+import software.amazon.awssdk.services.dsql.model.GetClusterResponse;
 
-import java.util.HashMap;
+import java.time.Duration;
 import java.util.Map;
 
 public class CreateCluster {
 
-    public static void main(String[] args) throws Exception {
+    public static void main(String[] args) {
```
imforster, line 16: nit: style should we have main at bottom instead?

java/cluster_management/src/main/java/org/example/DeleteCluster.java
```diff
@@ -5,45 +5,42 @@
...
-                .overrideConfiguration(clientOverrideConfiguration)
-                .region(region)
-                .credentialsProvider(DefaultCredentialsProvider.create())
-                .build();
-
-        String cluster_id = "foo0bar1baz2quux3quuux4";
-
-        DeleteClusterResponse response = deleteCluster(cluster_id, client);
-        System.out.println("Deleting Cluster with ID: " + cluster_id + ", Status: " + response.status());
+        String clusterId = "<your cluster id>";
```
imforster, line 20: So we discussed in our standup and think we should support using Environment variables to allow for github integration.

java/cluster_management/src/main/java/org/example/GetCluster.java
```diff
@@ -1,49 +1,28 @@
...
-                .overrideConfiguration(clientOverrideConfiguration)
-                .region(region)
-                .credentialsProvider(DefaultCredentialsProvider.create())
-                .build();
-
-        String cluster_id = "foo0bar1baz2quux3quuux4";
-
-        GetClusterResponse response = getCluster(cluster_id, client);
-        System.out.println("cluster status: " + response.status());
+        String clusterId = "<your cluster id>";
```
imforster, line 12: Env?

java/cluster_management/src/main/java/org/example/UpdateCluster.java
```diff
@@ -1,47 +1,31 @@
...
-                .region(region)
-                .credentialsProvider(DefaultCredentialsProvider.create())
-                .build();
-
-        String cluster_id = "foo0bar1baz2quux3quuux4";
-        Boolean deletionProtectionEnabled = false;
-
-        UpdateClusterResponse response = updateCluster(cluster_id, deletionProtectionEnabled, client);
-        System.out.println("Deletion Protection updating to: " + deletionProtectionEnabled.toString() + ", Status: " + response.status());
+        String clusterId = "<your cluster id>";
```
imforster, line 12: Env?

java/cluster_management/src/test/java/org/example/DsqlClusterManagementTest.java
```diff
@@ -1,92 +1,125 @@
...
+
+    /**
+     * Delete all clusters that are:
+     * <ol>
+     *     <li>Not already deleting; and,</li>
+     *     <li>Tagged with 'Repo=aws-samples/aurora-dsql-samples'; and,</li>
+     *     <li>Tagged with 'Name=java *'</li>
+     * </ol>
+     */
+    static void deleteTestsClusters(DsqlClient client) {
```
imforster, line 88: When we worked on HowTos in past we discovered that the github can interrupt so this should help resolve orphan clusters (Only caveat is if we have already reached our cluster limit we will get an ugly error first).

java/cluster_management/src/main/java/org/example/DeleteCluster.java
trstephen-amazon: And provide instructions for how to run each example by itself? I think that'll get really messy as we explain how to invoke each example, and the examples are simple enough to be self-documenting.

The structure of these files falls out of satisfying both 1/ a standalone example of how to invoke an API (parity with published doc samples; a single place to reference for dependency setup and input parameters) 2/ validation through an integration test.

java/cluster_management/src/main/java/org/example/CreateCluster.java
trstephen-amazon: As is it reads imperatively: marshal input params -> set up client -> invoke. Other languages may have stricter requirements about the order of declared methods (i.e. no use before declare) but I'm biasing for readability.

java/cluster_management/src/main/java/org/example/UpdateCluster.java
trstephen-amazon: implemented in next rev

java/cluster_management/src/main/java/org/example/GetCluster.java
trstephen-amazon: implemented in next rev

java/cluster_management/src/main/java/org/example/DeleteCluster.java
trstephen-amazon: implemented in next rev

found a way to add env vars that's not too cluttered. Invocation of each operation is explained in the readme for motivated users but the expectation is you just run `mvn test`

java/cluster_management/src/test/java/org/example/DsqlClusterManagementTest.java
danielfrankcom, line 88: Does sweeping for clusters belong in the test suite for each language, or should we have a single implementation somewhere which can be run against any account for any language?

As it stands it seems like we will have to implement similar logic to this in the other languages, which seems like an implementation and maintenance burden which we can avoid. Could this be separated out so we can reuse it for other languages, and execute it as a separate job at the end of the workflow?

trstephen-amazon: I suppose we could put a shell script or similar in this repo that can clean handle clean up. Each cluster_management workflow could add it at the end and it'd discover creds from earlier workflow steps.

It's a good idea! But I'm not going to pull it in scope here. Would need some additional work to align other examples on a discovery process (probably easy, use the 'Repo' tag like I do here). I'll make an Issue for this as a reminder.

java/cluster_management/README.md
```diff
@@ -2,23 +2,65 @@
 
 ## Overview
 
-The code examples in this topic show you how to use the Java PgJDBC to work with Aurora DSQL.
+The code examples in this topic show you how to use the AWS Java SDK v2 with DSQL
+to create, update, get, and delete single- and multi-Region clusters.
+
+Each file in the [/example](src/main/java/org/example) directory demonstrates a minimum
```
mitchell-elholm, line 8: Nit: minimum -> minimal

java/cluster_management/README.md
```diff
@@ -2,23 +2,65 @@
...
 
-- java version >= 17 is needed
+- java version >= 17 is installed.
+- Valid AWS credentials can be discovered by the [default provider chain](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-chain.html).
 
-### Run the example tests
+### Execute tests to create and delete clusters
+
+See [`DsqlClusterManagementTest.java`](src/test/java/org/example/DsqlClusterManagementTest.java) for configuration
+used in the `setup()` function.
```
mitchell-elholm, line 31: Nit: Considering the configuration is pretty much just the regions, it might be clearer to just state what's being used (ie. Uses environment properties with defaults of us-east-1/2/west-2).

mitchell-elholm (approved):

java/cluster_management/README.md
trstephen-amazon: I think I can make due with just the inline comments in the code fence

trstephen-amazon:
Offline: got OK to merge from @imforster and @danielfrankcom

PR #135, opened 2025-05-14
Python SR/MR cluster management

mitchell-elholm:
- Modifies Python cluster management to incorporate new API and to match documentation.
- Adds test cluster cleanup


python/cluster_management/test/test_dsql_cluster_management.py
```diff
@@ -0,0 +1,120 @@
...
+    except Exception as e:
+        pytest.fail(f"Unexpected exception: {e}")
+
+
+def delete_tests_clusters(region):
+    """
+    Delete all clusters that are:
+        1. Not already deleting; and,
+        2. Tagged with 'Repo=aws-samples/aurora-dsql-samples'; and,
+        3. Tagged with 'Name=Python*'
```
trstephen-amazon, line 73: not blocking: could even be a bit stricter here and check `Python-CM-Example-*`

trstephen-amazon (commented):
Overall LGTM but I guess CI found some issues with dict lookups in f-strings (thank you CI!)

trstephen-amazon:
aha, issue with the original commit was the f-string escapes are only supported in 3.12 ([docs](https://docs.python.org/3.12/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings), h/t @danielfrankcom )

Our CI is [using 3.10](https://github.com/aws-samples/aurora-dsql-samples/blob/d758abedd53806aa8dbedf98a2244d725747a6ee/.github/workflows/python-cm-integ-tests.yml#L40) and the [readme wants Python >= 3.8](https://github.com/aws-samples/aurora-dsql-samples/blob/d758abedd53806aa8dbedf98a2244d725747a6ee/python/cluster_management/README.md?plain=1#L25) so we need to rationalize the code, readme, and CI on a min version.

danielfrankcom:
A possible way to determine a sensible Python version would be to figure out the minimum version supported by all Python drivers which we have samples for. I know Psycopg3 had a clear list of supported/tested Python versions when I was working on that sample.

Using a single Python version across all Python samples in the project would help us keep track of these kind of issues.

python/cluster_management/test/test_dsql_cluster_management.py
```diff
@@ -0,0 +1,120 @@
...
+    for page in paginator.paginate():
+        for listed_cluster in page['clusters']:
+            # Get detailed cluster info
+            cluster = client.get_cluster(identifier=listed_cluster['identifier'])
+
+            # Skip clusters that are already being deleted or are deleted
+            if cluster['status'] in ['DELETED', 'DELETING']:
+                continue
+
+            # Check tags to identify test clusters
```
trstephen-amazon, line 92: not blocking: the updated SDK should already have tags in the GetCluster response

https://docs.aws.amazon.com/aurora-dsql/latest/APIReference/API_GetCluster.html

mitchell-elholm:
> A possible way to determine a sensible Python version would be to figure out the minimum version supported by all Python drivers which we have samples for. I know Psycopg3 had a clear list of supported/tested Python versions when I was working on that sample.
> 
> Using a single Python version across all Python samples in the project would help us keep track of these kinds of issues.

The latest version of Psycopg3 requires minimum 3.10, I don't remember any discussion about this but I'm guessing it's not a coincidence that we chose 3.10 in the CI. Psycopg2 only requires 3.8 though.

trstephen-amazon (approved):
🚢

PR #137, opened 2025-05-14
rust: update cluster management examples to use new API

danielfrankcom:
This change updates the Rust cluster management examples as follows:

- Use new AWS SDK for multi-region management
- Standardize examples based on AWS docs
- Clean up documentation
- Separate standalone binaries from tests
- Add formatting check to workflow
- Add binary execution check to workflow


.github/workflows/rust-cm-integ-tests.yml
```diff
@@ -20,6 +20,23 @@ on:
...
+  format:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Rust
+        run: rustup update
```
trstephen-amazon, line 32: kinda weird we don't pin a version here, but rust releases new versions every 6wk or so and that'd be a pain to maintain.

.github/workflows/rust-cm-integ-tests.yml
```diff
@@ -20,6 +20,23 @@ on:
...
+
+      - name: Check formatting
+        working-directory: ./rust/cluster_management
+        run: |
+          rustup component add rustfmt
+          cargo fmt -- --check
```
trstephen-amazon, line 38: thank you!

IME this can be a blocker for rust version bumps as new checks are introduced (see above comment about not pinning the rust version) but I'm willing to see if it becomes an annoyance. This code base is pretty small so it probably won't get tripped up by _every_ new version.

.github/workflows/rust-cm-integ-tests.yml
```diff
@@ -32,19 +49,49 @@ jobs:
...
+          role-to-assume: ${{ secrets.RUST_IAM_ROLE }}
+          aws-region: us-east-1
+
+      - name: Configure and run integration for cluster management
+        working-directory: ./rust/cluster_management
+        run: |
+          cargo test -- --nocapture
+
+  # This is a sanity check to make sure the examples are independently executable. We don't need the binary to do
+  # anything useful here, beyond running long enough to complain about the missing environment variable.
```
trstephen-amazon, line 70: Not sure I understand what we're guarding against here since we only check one. I'll have another think after reading through the examples but this'd probably be fine to leave out.

---

Finished reading: I still don't think this is necessary but I don't feel strong. Not blocking, but I'd break on the side of 'exclude' because CI ain't free

rust/cluster_management/src/bin/create_multi_region_clusters.rs
```diff
@@ -0,0 +1,139 @@
...
+    let config = Config::builder()
+        .behavior_version(BehaviorVersion::latest())
+        .credentials_provider(credentials)
+        .region(Region::new(region.to_owned()))
+        .build();
+
+    Client::from_conf(config)
+}
+
+/// Create a cluster without delete protection and a name
```
trstephen-amazon, line 27: > without deletion protection

But it's `true` on L52?

rust/cluster_management/src/bin/create_single_region_cluster.rs
```diff
@@ -0,0 +1,70 @@
...
+    let config = Config::builder()
+        .behavior_version(BehaviorVersion::latest())
+        .credentials_provider(credentials)
+        .region(Region::new(region.to_owned()))
+        .build();
+
+    Client::from_conf(config)
+}
+
+/// Create a cluster without delete protection and a name
```
trstephen-amazon, line 29: also sets deletion protection on L46

rust/cluster_management/src/bin/create_single_region_cluster.rs
```diff
@@ -0,0 +1,70 @@
...
+        ),
+    ]);
+
+    let create_cluster_output = client
+        .create_cluster()
+        .set_tags(Some(tags))
+        .deletion_protection_enabled(true)
+        .send()
+        .await
+        .unwrap();
```
trstephen-amazon, line 49: optional, not blocking: since we're already using `anyhow` in main you could  have this function return `anyhow::Result<GetClusterOutput>` and use `?` here instead of unwrapping.

trstephen-amazon (approved):
No blockers. Left some optional points you can address at your discretion, or merge.

.github/workflows/rust-cm-integ-tests.yml
danielfrankcom: I wonder if that's something dependabot can help with?

.github/workflows/rust-cm-integ-tests.yml
danielfrankcom: The main thing I'm guarding against here is any modifications that prevent the examples from building by themselves. There's a bit of a split in the way the example files are used, since they are each standalone binaries with a `main` method, as well as providing a library for the unit tests to use.

The unit tests are mostly for our benefit so CI can verify the code is correct, but I would think most customers will want to run the examples directly using the `main` entry point.

The CI job here is intended to make sure we don't accidentally drop that use case, and I'm only running one example since any accident that breaks this will likely affect all examples. Now that I'm thinking about it again though, the build is already done by this point in the CI job so we could just run all examples the same way I'm running the current one and have it fail since it should take no extra time to do that.

rust/cluster_management/src/bin/create_multi_region_clusters.rs
danielfrankcom: Good catch, will fix

.github/workflows/rust-cm-integ-tests.yml
danielfrankcom: Looks like there's an open issue for this but it's not possible right now [ref](https://github.com/dependabot/dependabot-core/issues/1702)

We could create a toolchain file if we wanted to use pinned versions, but as you say that may be a pain to keep updated manually.

.github/workflows/rust-cm-integ-tests.yml
danielfrankcom: This comment reminded me that I didn't run `clippy` against this code, which I just did locally and fixed one warning. In an ideal world we would have a `clippy` check job too, but I think that would be even more affected by the Rust version than the formatting would.

trstephen-amazon: Ah, right, most of the friction on rust version upgrade come from `clippy`, not `fmt`.

rust/cluster_management/src/bin/create_single_region_cluster.rs
danielfrankcom: Agreed, there are quite a few `unwrap` calls throughout the examples which don't make a ton of sense. Next push will clean them up.

danielfrankcom: Should be addressed by [f4af901](https://github.com/aws-samples/aurora-dsql-samples/commit/f4af901e820d64d7e859b67c9b22d16a29713f19)

rust/cluster_management/src/bin/create_single_region_cluster.rs
danielfrankcom: Should be addressed by [f4af901](https://github.com/aws-samples/aurora-dsql-samples/commit/f4af901e820d64d7e859b67c9b22d16a29713f19)

rust/cluster_management/src/bin/create_multi_region_clusters.rs
danielfrankcom: Should be addressed by [f4af901](https://github.com/aws-samples/aurora-dsql-samples/commit/f4af901e820d64d7e859b67c9b22d16a29713f19)

danielfrankcom:
- [f4af901](https://github.com/aws-samples/aurora-dsql-samples/commit/f4af901e820d64d7e859b67c9b22d16a29713f19) addresses most active comments.
- [1f86c9a](https://github.com/aws-samples/aurora-dsql-samples/commit/1f86c9a8e258c9eb8a46ae4145fc9c926d49ad88) updates the CI job to check all standalone binaries, and increases the job timeout for the test run from 15 mins to 30 due to a recent failure.

danielfrankcom:
- [49e66f6](https://github.com/aws-samples/aurora-dsql-samples/commit/49e66f67e5f04e00b1337a1484f0108d740fa03a) makes the binary check job more permissive about the env var which is checked for

.github/workflows/rust-cm-integ-tests.yml
danielfrankcom: I've updated the job to run all of the binaries, since the overhead on top of running 1 binary is negligible. I've also updating the wording of the comment to hopefully make it a bit clearer why it is here.

.github/workflows/rust-cm-integ-tests.yml
```diff
@@ -20,9 +20,26 @@ on:
...
+      - name: Check formatting
+        working-directory: ./rust/cluster_management
+        run: |
+          rustup component add rustfmt
+          cargo fmt -- --check
+
   test:
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 30
```
trstephen-amazon, line 42: just paranoia or does it actually take >15m to run? (CI is at 11m27s)

.github/workflows/rust-cm-integ-tests.yml
```diff
@@ -32,19 +49,58 @@ jobs:
...
+          
+          # Test each binary
+          for bin in $binaries; do
+            echo "Testing binary: $bin"
+
+            # Run the binary and capture its output
+            output=$(cargo run --bin $bin 2>&1) || true
+
+            # Check if the output contains the expected error message about environment variables
+            if echo "$output" | grep -q "env variable \`.*\` should be set"; then
```
trstephen-amazon, line 98: This makes it really hard to derive any conclusions about the binaries beyond 'they exist' and 'they want an environment variable'. Does it expect the right environment variable? Does it expect the right amount of variables? Does it use them correctly? And so on.

From https://github.com/aws-samples/aurora-dsql-samples/pull/137#discussion_r2090079211 :
> The CI job here is intended to make sure we don't accidentally drop [the binary invocation] use case

Wouldn't the build system catch this? Or are you concerned about a change that removes binary creation _and_ alters the build in the same commit? If that's what we're guarding against then I think you could also communicate "don't change the package structure" with a comment in `Cargo.toml`.

rust/cluster_management/src/bin/create_single_region_cluster.rs
```diff
@@ -0,0 +1,70 @@
...
+    println!("Created {}", create_cluster_output.arn);
+
+    println!("Waiting for cluster to become ACTIVE");
+    client
+        .wait_until_cluster_active()
+        .identifier(&create_cluster_output.identifier)
+        .wait(std::time::Duration::from_secs(300)) // Wait up to 5 minutes
+        .await?
+        .into_result()
+        .map_err(Into::into)
```
trstephen-amazon, line 60: Not blocking: `.map_err(Into::into)` seems ... mysterious?

Consider:
```rust
let cluster = client...into_result()?;
Ok(cluster)
```

rust/cluster_management/tests/multi_region_test.rs
```diff
@@ -0,0 +1,68 @@
...
+
+        println!("Getting cluster details...");
+        let cluster_1_details = get_cluster(&region_1, &cluster_1_id).await?;
+        let cluster_2_details = get_cluster(&region_2, &cluster_2_id).await?;
+
+        let cluster_1_multi_region_props = cluster_1_details
+            .multi_region_properties
+            .ok_or_else(|| anyhow::anyhow!("No multi-region properties"))?;
+
+        let cluster_1_linked_clusters = cluster_1_multi_region_props
```
trstephen-amazon, line 32: "peered" is the updated verbiage for clusters, replacing "linked"

rust/cluster_management/tests/multi_region_test.rs
```diff
@@ -0,0 +1,68 @@
...
+            .clusters
+            .ok_or_else(|| anyhow::anyhow!("No linked clusters"))?;
+        assert!(cluster_1_linked_clusters.contains(&cluster_2_details.arn));
+
+        let cluster_1_witness_region = cluster_1_multi_region_props
+            .witness_region
+            .ok_or_else(|| anyhow::anyhow!("No witness region"))?;
+        assert!(cluster_1_witness_region.eq(&witness_region));
```
trstephen-amazon, line 40: IMO it's fine to just assert the cluster became ACTIVE. That implies the multi region props were set correctly.

rust/cluster_management/tests/multi_region_test.rs
```diff
@@ -0,0 +1,68 @@
...
+        assert!(cluster_2_linked_clusters.contains(&cluster_1_details.arn));
+
+        let cluster_2_witness_region = cluster_2_multi_region_props
+            .witness_region
+            .ok_or_else(|| anyhow::anyhow!("No witness region"))?;
+        assert!(cluster_2_witness_region.eq(&witness_region));
+
+        println!("Updating clusters to remove deletion protection...");
+        let update_1_response = update_cluster(&region_1, &cluster_1_id).await?;
+        assert_eq!(update_1_response.identifier, cluster_1_id);
```
trstephen-amazon, line 58: eh, seems like a trivial assertion. It's not _wrong_ but to figure out what this test is doing I have to visually filter out quite a bit.

trstephen-amazon (approved):
I'm OK with this as-is. Still not convinced of the binary poking in CI but it's very unlikely to be a burden.

rust/cluster_management/tests/multi_region_test.rs
danielfrankcom: Should be addressed in the next push. There were quite a few "linked" references which I've updated.

rust/cluster_management/tests/multi_region_test.rs
danielfrankcom: Should be addressed in the next push. I think you're right, I've removed this whole section checking the properties in detail and am now just checking `ACTIVE`.

rust/cluster_management/tests/multi_region_test.rs
danielfrankcom: I've replaced this with a check for the `UPDATING` state which should hopefully be less trivial and relevant to the action above.

rust/cluster_management/src/bin/create_single_region_cluster.rs
danielfrankcom: Thanks, this is cleaner

.github/workflows/rust-cm-integ-tests.yml
danielfrankcom: I've come around on this. I did some testing and the tests won't even run if one of the `main` methods are removed since the binary crate gets run as part of the test compilation. I will remove this job as I agree it's not necessary, and add a comment in the `Cargo.toml` accordingly.

PR #143, opened 2025-05-16
Add Postgres-js CI

mitchell-elholm:
- Adds an integration test workflow for Postgres-js

### DIFF
--- a/.github/workflows/javascript-postgresjs-integ-tests.yml
+++ b/.github/workflows/javascript-postgresjs-integ-tests.yml
@@ -1,0 +1,56 @@
+name: Postgres-js integration tests
+
+on:
+  push:
+    branches: [ main ]
+    paths:
+      - 'javascript/postgres-js/**'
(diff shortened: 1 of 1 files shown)


imforster (approved):

.github/workflows/javascript-postgresjs-integ-tests.yml
```diff
@@ -0,0 +1,56 @@
...
+      pull-requests: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18.x'
```
wcmjunior, line 40: We should use node `22.x` instead as `18.x` is reaching end of life.

wcmjunior, line 40: We should use node `22.x` instead as `18.x` is reaching end of life.

wcmjunior (changes requested):
See

wcmjunior (approved):

PR #144, opened 2025-05-16
Clean up lingering clusters after test runs

### LINKED ISSUE #136: Portable cluster_management cleanup
> Does sweeping for clusters belong in the test suite for each language, or should we have a single implementation somewhere which can be run against any account for any language?
>
> As it stands it seems like we will have to implement similar logic to this in the other languages, which seems like an implementation and maintenance burden which we can avoid. Could this be separated out so we can reuse it for other languages, and execute it as a separate job at the end of the workflow?

_Originally posted by @danielfrankcom in https://github.com/aws-samples/aurora-dsql-samples/pull/128#discussion_r2089861812_

---

Make a simple, portable script that can be added to any cluster_management example to delete clusters tagged with `Name=aws-samples/aurora-dsql-samples`

danielfrankcom:
This PR fixes #136 

This change introduces a script to clean up lingering clusters. It only applies to clusters that are not already deleting, and which have the correct `Repo` tag. The script also exits early if the `IS_CI` flag isn't set for additional safety.

The `clean-clusters.yml` workflow can be called to clean up the clusters generated by other workflows.o

Things needed before review:

- [x] Use latest AWS CLI version. Current one in workflow isn't showing the tags properly which breaks the script. ~~Maybe add a sanity check for the minimum version to the script.~~ This is a bit complicated in Bash so I won't bother for now
- [x] Modify other workflow files to reference this shared implementation
- [x] Remove duplicate implementations in language test suites

### DIFF
--- a/.github/scripts/clean-clusters.sh
+++ b/.github/scripts/clean-clusters.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+
+if [ -z "${IS_CI}" ]; then
+  echo "Error: This script is intended to run only in CI environments."
+  echo "Running it locally may delete clusters in your account."
+  echo "Set the IS_CI environment variable to run this script."
+  exit 1
+fi
+
+REGIONS=("us-east-1" "us-east-2")
+
+# Region and cluster ID can be extracted from ARN
+# ARN format: arn:aws:dsql:REGION:ACCOUNT:cluster/CLUSTER_ID
+declare -a ARNS=()
+declare -a FILTERED_ARNS=()
+
+# Get clusters from each region and extract ARNs
+for region in "${REGIONS[@]}"; do
+  echo "Listing clusters in $region..."
+
+  region_arns=$(aws dsql list-clusters --region "$region" | jq -r '.clusters[].arn')
+
+  # Add ARNs to the array if any were found
+  if [ -n "$region_arns" ]; then
+    while IFS= read -r arn; do
+      ARNS+=("$arn")
+    done <<< "$region_arns"
+  fi
+done
+
+echo -e "\nFound ${#ARNS[@]} cluster(s) across all regions:"
+printf '%s\n' "${ARNS[@]}"
+
+echo -e "\nFiltering clusters..."
+for arn in "${ARNS[@]}"; do
+  region=$(echo "$arn" | cut -d':' -f4)
+  cluster_id=$(echo "$arn" | cut -d'/' -f2)
+
+  echo "Checking cluster $cluster_id in region $region..."
+
+  cluster_details=$(aws dsql get-cluster --region "$region" --identifier "$cluster_id")
+
+  status=$(echo "$cluster_details" | jq -r '.status')
+  repo_tag=$(echo "$cluster_details" | jq -r '.tags.Repo // empty')
+
+  # We only want clusters that are not already deleting, and have the specific repo tag
+  if [[ "$status" != "DELETED" && "$status" != "DELETING" && "$repo_tag" == "aws-samples/aurora-dsql-samples" ]]; then
+    echo "Cluster $cluster_id qualifies for update (Status: $status, Repo tag: $repo_tag)"
+    FILTERED_ARNS+=("$arn")
+  else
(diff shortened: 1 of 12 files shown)


mitchell-elholm:
Will this include a name parameter to filter clusters to delete? Or will it just be the repo tag?

danielfrankcom:
> Will this include a name parameter to filter clusters to delete? Or will it just be the repo tag?

My intention is just the repo tag. We could add a name parameter if we wanted but I'm not sure it's necessary with the way the examples are defined at the moment. It would also be a bit complicated for Go since the name is dynamic and includes the workflow run ID or something.

danielfrankcom:
[This](https://github.com/aws-samples/aurora-dsql-samples/actions/runs/15076454160/job/42385560942) is a successful run of the cleanup script which shows what it does when actually deleting things. I stacked the Rust account with various types and states of leftover clusters before this run to give the script something to do.

.github/scripts/clean-clusters.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+
+if [ -z "${IS_CI}" ]; then
+  echo "Error: This script is intended to run only in CI environments."
+  echo "Running it in your account may have destructive effects."
```
mitchell-elholm, line 5: Nit: Maybe clarify that "destructive effects" means deleting clusters.

.github/workflows/python-cm-integ-tests.yml
```diff
@@ -67,4 +67,15 @@ jobs:
         pip list
         echo "$GITHUB_WORKSPACE" >> $GITHUB_PATH
         pytest -v test/
-        
+
```
mitchell-elholm, line 70: Should also delete the IS_CI env variables from Java, Python, and any others.

.github/scripts/clean-clusters.sh
```diff
@@ -0,0 +1,82 @@
...
+  echo "Set the IS_CI environment variable to run this script."
+  exit 1
+fi
+
+REGIONS=("us-east-1" "us-east-2")
```
mitchell-elholm, line 10: I don't think we need to change this right now, but I think we should probably move CRUD regions to Github secrets/variables to avoid duplication between this file and the examples themselves.

.github/workflows/python-cm-integ-tests.yml
danielfrankcom: Oh good catch, I missed that. I think only Java and Python ended up merging a version of this but I'll check for other instances of this variable.

.github/scripts/clean-clusters.sh
danielfrankcom: Agreed. Won't touch this now but it should be easy to pull these from an environment variable instead.

.github/workflows/cpp-cm-integ-tests.yml
```diff
@@ -77,4 +77,16 @@ jobs:
...
         WITNESS_REGION: us-west-2
       run: |
-        ./example
\ No newline at end of file
+        ./example
+
+  cleanup:
+    if: always()
+    needs: cpp-cm-integ-test
+    uses: ./.github/workflows/clean-clusters.yml
```
trstephen-amazon, line 85: nice!

trstephen-amazon (approved):

.github/scripts/clean-clusters.sh
danielfrankcom: Created #146

.github/scripts/clean-clusters.sh
danielfrankcom: Should be addressed by [da6531e](https://github.com/aws-samples/aurora-dsql-samples/commit/da6531ef2c0a994fbef5f257805b8ea9322de7f1)

.github/workflows/python-cm-integ-tests.yml
danielfrankcom: Should be addressed by [da6531e](https://github.com/aws-samples/aurora-dsql-samples/commit/da6531ef2c0a994fbef5f257805b8ea9322de7f1)

mitchell-elholm (approved):

PR #147, opened 2025-05-17
Align all env variables for cluster management in all languages

vic-tsang:
- renamed all env variables to use the standard ones 

Single Cluster 
- CLUSTER_REGION
- CLUSTER_ID

Multi Clusters 
- CLUSTER_1_REGION  
- CLUSTER_2_REGION  
- WITNESS_REGION
- CLUSTER_1_ID
- CLUSTER_2_ID

### DIFF
--- a/go/cluster_management/cmd/delete_multi_region/delete_multi_region_cluster_integ_test.go
+++ b/go/cluster_management/cmd/delete_multi_region/delete_multi_region_cluster_integ_test.go
@@ -45,9 +45,9 @@ func setup() {
 		return
 	}
 
-	os.Setenv("REGION", "us-east-1")
+	os.Setenv("CLUSTER_1_REGION", "us-east-1")
 	os.Setenv("CLUSTER_1_ID", *output.Identifier)
-	os.Setenv("REGION2", "us-east-2")
+	os.Setenv("CLUSTER_2_REGION", "us-east-2")
 	os.Setenv("CLUSTER_2_ID", *output1.Identifier)
 }
 
--- a/go/cluster_management/cmd/delete_multi_region/delete_multi_region_cluster_integ_test.go
+++ b/go/cluster_management/cmd/delete_multi_region/delete_multi_region_cluster_integ_test.go
@@ -46,9 +46,9 @@ func setup() {
 	}
 
 	os.Setenv("REGION", "us-east-1")
-	os.Setenv("CLUSTER1_ID", *output.Identifier)
+	os.Setenv("CLUSTER_1_ID", *output.Identifier)
 	os.Setenv("REGION2", "us-east-2")
-	os.Setenv("CLUSTER2_ID", *output1.Identifier)
+	os.Setenv("CLUSTER_2_ID", *output1.Identifier)
 }
 
 func teardown() {
@@ -69,9 +69,9 @@ func TestDeleteMultiRegionClustersRegion(t *testing.T) {
 		{
 			name:        "Delete multi-region cluster",
 			region1:     "us-east-1",
-			identifier1: os.Getenv("CLUSTER1_ID"),
+			identifier1: os.Getenv("CLUSTER_1_ID"),
 			region2:     "us-east-2",
-			identifier2: os.Getenv("CLUSTER2_ID"),
+			identifier2: os.Getenv("CLUSTER_2_ID"),
 			wantErr:     false,
 		},
 	}
--- a/python/cluster_management/README.md
+++ b/python/cluster_management/README.md
@@ -13,11 +13,11 @@ in [`test_dsql_cluster_management.py`](test/test_dsql_cluster_management.py).
 
 ### ⚠️ Important
 
-* Running this code might result in charges to your AWS account.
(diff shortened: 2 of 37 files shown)


python/cluster_management/src/delete_single_region.py
```diff
@@ -22,9 +22,9 @@ def delete_cluster(region, identifier):
 
 
 def main():
-    region = os.environ.get("REGION_1", "us-east-1")
-    cluster_id = os.environ.get("CLUSTER_ID_1")
-    assert cluster_id is not None, "Must provide CLUSTER_ID_1"
+    region = os.environ.get("CLUSTER_1_REGION", "us-east-1")
```
danielfrankcom, line 25: Given this is a single region test I think this should be `CLUSTER_REGION`

danielfrankcom: Same with `CLUSTER_ID`

python/cluster_management/src/get_cluster.py
```diff
@@ -14,9 +14,9 @@ def get_cluster(region, identifier):
 
 
 def main():
-    region = os.environ.get("REGION_1", "us-east-1")
-    cluster_id = os.environ.get("CLUSTER_ID_1")
-    assert cluster_id is not None, "Must provide CLUSTER_ID_1"
+    region = os.environ.get("CLUSTER_1_REGION", "us-east-1")
```
danielfrankcom, line 17: Should be `CLUSTER_REGION` and `CLUSTER_ID`

python/cluster_management/src/update_cluster.py
```diff
@@ -12,9 +12,9 @@ def update_cluster(region, cluster_id, deletion_protection_enabled):
 
 
 def main():
-    region = os.environ.get("REGION_1", "us-east-1")
-    cluster_id = os.environ.get("CLUSTER_ID_1")
-    assert cluster_id is not None, "Must provide CLUSTER_ID_1"
+    region = os.environ.get("CLUSTER_1_REGION", "us-east-1")
```
danielfrankcom, line 15: Should be `CLUSTER_REGION` and `CLUSTER_ID`

python/cluster_management/README.md
```diff
@@ -47,16 +47,17 @@ pytest test/test_dsql_cluster_management.py [-s]
 Test execution will take around five minutes as it waits for clusters to complete activation and deletion.
 
 ### Executing single operations
+
 Files in [src/](src/) each have a `main()` function that let you exercise single operations.
 
 ```shell
 # Check each operation for its expected environment variables
-REGION_1="us-east-1" CLUSTER_ID_1="<your cluster id>" \
+CLUSTER_1_REGION="us-east-1" CLUSTER_1_ID="<your cluster id>" \
```
danielfrankcom, line 55: If we change the `get_cluster.py` file then this will probably need to be updated to match.

go/cluster_management/cmd/delete_multi_region/delete_multi_region_cluster_integ_test.go
```diff
@@ -46,9 +46,9 @@ func setup() {
 	}
 
 	os.Setenv("REGION", "us-east-1")
```
danielfrankcom, line 48: Should this be `CLUSTER_1_REGION` and `CLUSTER_2_REGION`?

danielfrankcom (changes requested):
Changes here mostly look good.

I did a quick search in the branch and found the following which I think were missed:

- `go/cluster_management/cmd/update_cluster/update_cluster_integ_test.go`

```
			region:     os.Getenv("REGION"),
			identifier: os.Getenv("CLUSTER_ID"),
```

- `go/cluster_management/cmd/get_cluster/get_cluster_integ_test.go`

```
			region:     os.Getenv("REGION"),
			identifier: os.Getenv("CLUSTER_ID"),
```

For whatever reason it seems these are also setting environment variables but any change to that is likely out of scope. We should use consistent names imo. For extra weirdness points, the `create_multi_region_integ_test.go` file and the single-region equivalent don't use environment variables at all. Even the ones I mentioned above inconsistently reference the region, sometimes with an env var and sometimes hardcoded.

danielfrankcom (approved):

danielfrankcom (approved):

python/cluster_management/test/test_dsql_cluster_management.py
```diff
@@ -9,8 +9,8 @@
 import os
 import pytest
 
-region_1 = os.environ.get("REGION_1", "us-east-1")
```
mitchell-elholm, line 12: We should be using CLUSTER_REGION for the single region clusters here. Can change the usages of region_1 in the single region test to a new region var.

mitchell-elholm (changes requested):

python/cluster_management/test/test_dsql_cluster_management.py
vic-tsang: fixed

mitchell-elholm (approved):

danielfrankcom (approved):

PR #157, opened 2025-05-28
cpp: Run cpp workflows faster

danielfrankcom:
This PR updates the C++ workflows to improve the runtime.

The following changes are included:
- Run the build in parallel given the number of processors allocated to the worker
- Disable tests in third-party dependencies

This improves the time it takes to install the AWS SDK from ~8:30 to ~3 minutes.

These changes will have less of an effect on the cluster management tests since the test run there takes a significant amount of time by itself. It will have more of an effect on the `libpq` tests since the test runtime for those is near instant.


leszek-bq (approved):
Looks good to me.

PR #160, opened 2025-05-28
java: Use direct TLS connections for pgJDBC

danielfrankcom:
This PR modifies the `pgJDBC` example to set `sslNegotiation=direct` as per #150.

I added a section to the `README.md` file which describes more about the TLS configuration, and calls out that direct TLS connections are a new feature and may not be supported by all driver versions. My intention is to add the same section to the `README.md` files of other drivers as they are updated.

By submitting this pull request, I confirm that my contribution is made under the terms of the MIT-0 license.

Issue #150, opened 2025-05-23
Update Samples to use PG Direct TLS Negotiation


rmarrowstone:
DSQL supports Postgres' Direct TLS Negotiation. This is a best practice for performance. The samples should be updated to support this, if they are using a PG 17 compatible client.

For more: https://marc-bowes.com/postgres-direct-tls.html

danielfrankcom:
There are now PRs for each of the drivers which currently support direct TLS negotiation.

I created an issue (https://github.com/launchbadge/sqlx/issues/3880) and PR (https://github.com/launchbadge/sqlx/pull/3879) for SQLx. We should be able to update the sample if that change is accepted by the maintainers.

At the time of writing, the various JavaScript drivers are the only ones which we need upstream support for.

rmarrowstone:
Awesome! Thank you.

wcmjunior:
@rmarrowstone I asked @danielfrankcom to close this ticket once he finishes the changes for those drivers that currently support the direct TLS negotiation and open individual tickets for those that don't support. This will make it easier to track and fix in the future, otherwise this ticket will hang indefinitely.

@danielfrankcom I kindly ask you to relate all of the to this ticket so this can be referred in the future. Thanks!

wcmjunior:
Sub-issues is even better. :)

PR #167, opened 2025-05-29
python: Use direct TLS connections for psycopg2

danielfrankcom:
This PR modifies the `psycopg2` sample to set `sslnegotiation=direct` as per #150.

It does this conditionally based on the underlying `libpq` version, since the same `psycopg2` version can be used with multiple versions of `libpq`. I verified in the release notes [here](https://www.postgresql.org/docs/release/17.0/) that `sslnegotiation` has been available since the `17.0` release and was not added as a minor version bump.

At the time of writing, the `psycopg2-binary` package includes `libpq` 16.0, which doesn't support the `sslnegotiation` parameter. I manually checked a version of `psycopg2` compiled against `libpq` 17.4 and the direct TLS connection worked properly. I have kept the sample dependency on the pre-built binary, as it's more approachable in terms of the system requirement for running the sample. If/when `psycopg2-binary` is updated to include `libpq` 17.0+, it will start using `sslnegotiation` without any changes here.

I added a section to the `README.md` file which describes more about the TLS configuration, and calls out that direct TLS connections are a new feature and may not be supported by all driver versions. This section was already reviewed in #160, and will be added to all samples which support direct TLS connections.

### DIFF
--- a/python/psycopg2/README.md
+++ b/python/psycopg2/README.md
@@ -29,6 +29,19 @@ The code automatically detects the user type and adjusts its behavior accordingl
 * This code is not tested in every AWS Region. For more information, see
   [AWS Regional Services](https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services).
 
+## TLS connection configuration
+
+This example uses direct TLS connections where supported, and verifies the server certificate is trusted. Verified SSL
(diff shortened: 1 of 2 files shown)


leszek-bq (approved):
LGTM

PR #168, opened 2025-05-29
python: Use direct TLS connections for SQLAlchemy

danielfrankcom:
This PR modifies the `SQLAlchemy` sample to set `sslnegotiation=direct` as per #150.

It does this conditionally based on the underlying `libpq` version, since the same `psycopg2` version can be used with multiple versions of `libpq`. I verified in the release notes [here](https://www.postgresql.org/docs/release/17.0/) that `sslnegotiation` has been available since the `17.0` release and was not added as a minor version bump.

As with #167, at the time of writing, the `psycopg2-binary` package includes `libpq` 16.0, which doesn't support the `sslnegotiation` parameter. See the other PR for more detail, as the same applies here given SQLAlchemy's use of `psycopg2` in this sample.

I added a section to the `README.md` file which describes more about the TLS configuration, and calls out that direct TLS connections are a new feature and may not be supported by all driver versions. This section was already reviewed in #160, and will be added to all samples which support direct TLS connections.

### DIFF
--- a/python/sqlalchemy/README.md
+++ b/python/sqlalchemy/README.md
@@ -33,6 +33,19 @@ The code automatically detects the user type and adjusts its behavior accordingl
 * This code is not tested in every AWS Region. For more information, see
   [AWS Regional Services](https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services).
 
+## TLS connection configuration
+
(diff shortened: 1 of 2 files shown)


leszek-bq (approved):
LGTM

PR #179, opened 2025-06-19
pgx: Configure connections using hooks rather than replacing pool

danielfrankcom:
This PR modifies the Go pgx driver sample to resolve a race-condition with the way the pool was being handled.

The pgx pool was previously wrapped in a custom construct which replaced the inner pool when the token needed to be refreshed. This had a few issues:

- Established connections do not need to be replaced until the `MaxConnLifetime` expires, even if the token which was used to authenticate them has already expired.
- The token refresh logic replaced the pool, which meant any parties with a reference to the old pool would experience errors after the pool was closed.
- A mutex was used to reconfigure the pool wrapper, which means any safe access to perform operations on the pool should also acquire the mutex. This mutex acquisition was missing in the sample, and this approach also wouldn't scale to allow for highly concurrent use of the pool.

To address these concerns, I've removed the custom pool wrapped and utilized the `BeforeConnect` pool hook to generate an authentication token for each new connection on the fly. Since generating a token does not need to make any network requests, this is very fast, and has negligible overhead compared to opening a DB connection. This appears to be the recommended approach in https://github.com/jackc/pgx/issues/676.

Since this approach is fundamentally incompatible with the previous approach, the diff in this PR is quite large. There was a significant reduction in the size/complexity of this sample as part of this change, so I would recommend primarily focusing on the "new" side of the diff. I also updated the tests which relied pretty heavily on the structure of the previous approach.

### DIFF
--- a/go/pgx/example.go
+++ b/go/pgx/example.go
@@ -5,16 +5,14 @@ import (
 	"fmt"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/feature/dsql/auth"
-	"github.com/aws/aws-sdk-go-v2/service/dsql"
+	"github.com/google/uuid"
 	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"log"
 	"os"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
-
-	"github.com/google/uuid"
-	"github.com/jackc/pgx/v5/pgxpool"
 )
 
 type Owner struct {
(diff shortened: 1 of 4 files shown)


danielfrankcom:
Recent push fixes the workflow file which I forgot to update.

go/pgx/example.go
```diff
@@ -105,242 +106,103 @@ func getEnvInt(key string, defaultValue int) int {
...
-	// Generate initial token
-	token, err := GenerateDbConnectAuthToken(poolCtx, dbConfig.Host, dbConfig.Region, dbConfig.User)
-	if err != nil {
-		cancel()
-		return nil, fmt.Errorf("failed to generate auth token: %v", err)
-	}
+	// This doesn't need to be configurable for most applications, but we allow
+	// configuration here for the sake of unit testing.
+	expirySeconds := getEnvInt("TOKEN_EXPIRY_SECS", 30)
+	expiry := time.Duration(expirySeconds) * time.Second
```
imforster, line 128: Probably should call out that default token expiry interval is 15 min.

imforster (approved):

PR #185, opened 2025-07-15
Added pool options for TypeORM

vic-tsang:
Added the default pool settings for TypeORM for clarify of the example. 

### DIFF
--- a/typescript/type-orm/src/data-source.ts
+++ b/typescript/type-orm/src/data-source.ts
@@ -45,7 +45,7 @@ const getDataSource = async () => {
         max: 5,                      
         idleTimeoutMillis: 10000,     
         connectionTimeoutMillis: 30000,   
-        maxLifetimeSeconds: 3600,  // Maximum connection duration 1 hour (3600 seconds)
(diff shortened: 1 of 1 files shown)


typescript/type-orm/src/data-source.ts
```diff
@@ -39,6 +39,14 @@ const getDataSource = async () => {
       schema: schema,
       migrations: [path.join(__dirname, "/migrations/**/*{.ts,.js}")],
       migrationsRun: false,
+      // Pool options
+      extra: {
+        min: 0,
+        max: 5,                      
+        idleTimeoutMillis: 10000,     
+        connectionTimeoutMillis: 30000,   
+        maxLifetimeSeconds: 3600,  // Maximum connection duration 1 hour (3600 seconds)
```
danielfrankcom, line 48: Can we set this to a smaller value like 30 mins or 45? Currently this is the same as the connection lifetime limit which means the token may expire before the connection is evicted from the pool.

I think some of our older examples have the same problem, but we haven't gone back to fix it yet.

PR #194, opened 2025-10-03
Use package-lock.json and npm ci for predictable dependencies

danielfrankcom:
This PR ensures the dependencies described by the Node.js packages in the repo will use the same dependencies across all developer machines, including the workflows.

The change adds `package-lock.json` files where they were previously missing, and swaps the workflow files to use `npm ci` to avoid dependency versions being changed on the fly. In at least one case the `npm ci` failed when testing locally since the `package.json` and `package-lock.json` files were out of sync which should never happen on `main`.

### DIFF
--- a/.github/workflows/javascript-cm-integ-tests.yml
+++ b/.github/workflows/javascript-cm-integ-tests.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Configure and run integration for cluster management
         working-directory: ./javascript/cluster_management
         run: |
-          npm install
+          npm ci
           npm test
 
   cleanup:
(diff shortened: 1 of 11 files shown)


vic-tsang:
my question is what happens if we need to update the packages we need to update the package-lock.json ?

danielfrankcom:
> we need to update the package-lock.json ?

Yes, this is a standard part of the Node.js ecosystem. Previously we did not actually control the versions used for some of these samples, we just used a version based on the suggested range from the `package.json` file.

This change will let us make sure we are using the same version everywhere to prevent "works on my machine" issues.

vic-tsang (approved):

PR #197, opened 2025-10-07
java: Add sample for Spring Boot

danielfrankcom:
This PR adds a code sample for Spring Boot.

This replaces the sample which was removed in #196 with one that matches the style and approach taken by the existing Java samples. A GitHub workflow is also included to test this sample and prevent regressions.

The core of the sample is based on my change in https://github.com/aws/aws-advanced-jdbc-wrapper/pull/1481, but I tweaked things a fair bit to better match the existing Java samples. There are also obvious differences caused by the fact that this sample uses the Postgres driver directly without IAM wrapping.


danielfrankcom:
Secrets are added and workflows are passing now.

mitchell-elholm (approved):

PR #198, opened 2025-10-07
java: Use JDBC connector in pgJDBC sample

danielfrankcom:
This PR updates the pgJDBC sample to use the [Aurora DSQL JDBC Connector](https://github.com/awslabs/aurora-dsql-jdbc-connector).

This allows users to avoid making token generation SDK calls themselves, and simplifies the IAM auth experience.

### DIFF
--- a/java/pgjdbc/build.gradle.kts
+++ b/java/pgjdbc/build.gradle.kts
@@ -17,9 +17,6 @@ repositories {
 }
 
 dependencies {
-    implementation("org.postgresql:postgresql:42.7.8")
-    implementation("org.slf4j:slf4j-api:2.0.17")
-    implementation("org.slf4j:slf4j-simple:2.0.17")
     implementation("software.amazon.dsql:aurora-dsql-jdbc-connector:1.2.0")
 
     testImplementation(platform("org.junit:junit-bom:5.10.0"))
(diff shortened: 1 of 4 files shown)


danielfrankcom:
Pushed a chance to bump `org.postgresql:postgresql` from `42.7.4` to `42.7.8` to resolve a [security warning](https://github.com/advisories/GHSA-hq9p-pm7w-8p54).

java/pgjdbc/build.gradle.kts
```diff
@@ -17,13 +17,14 @@ repositories {
 }
 
 dependencies {
+    implementation("org.postgresql:postgresql:42.7.8")
```
mitchell-elholm, line 20: Doesn't the jdbc connector include pgJDBC as a transitive dependency? Or are we overriding the included driver version here?

danielfrankcom: The logging library was also not used by the example at all and is pulled in as a transitive dependency. Removed both dependencies and updated the Maven build command. The jar packaging approach which it used previously doesn't work with JDBC as a transitive dependency but I think that's fine since that's not how you would package a project typically anyway.

mitchell-elholm (approved):

PR #698, opened 2026-03-17
Add batch DELETE/UPDATE samples for datasets exceeding 3k row limit

rmconstantin:
Demonstrates sequential and parallel batch processing patterns for Aurora DSQL with OCC retry logic and recommended connection management. Includes Python (psycopg2), Java (pgJDBC), and Node.js (node-postgres) implementations.
Fixes #693 .


java/batch_operations/gradle/wrapper/gradle-wrapper.jar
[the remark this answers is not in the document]
rmconstantin: The `gradle-wrapper.jar` is the Gradle Wrapper bootstrap JAR. It allows anyone to build the project without having Gradle pre-installed — the wrapper downloads the correct Gradle version automatically. Shipping it in version control is the [recommended Gradle convention](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper). The other Java projects in this repo (`java/pgjdbc`, `java/spring_boot`) follow the same pattern.

I also updated `java/.gitignore` to stop ignoring these wrapper files (previously they were gitignored but force-tracked, which was inconsistent).

java/batch_operations/gradlew.bat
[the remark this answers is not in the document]
rmconstantin: Yes, `gradlew` and `gradlew.bat` should be checked in. They are the Gradle Wrapper scripts (Unix and Windows respectively) that bootstrap the build — users run `./gradlew build` instead of installing Gradle globally. This is the standard Gradle convention and matches `java/pgjdbc` and `java/spring_boot` in this repo.

I cleaned up `java/.gitignore` in the latest commit to remove the `**/gradlew`, `**/gradlew.bat`, and `**/gradle/` patterns that were incorrectly ignoring these files.

rmconstantin:
Rebased on main — conflicts resolved.

CI has two failures:

1. **`deps-review`** — Transitive npm dependencies from Jest (e.g. `color-name`, `co`, `ci-info`) score below the repo's OpenSSF Scorecard threshold of 3. These are all standard, widely-used packages pulled in by Jest. Is there an allow-list or policy exception for test dependencies?

2. **`javascript-node-postgres / create-cluster`** — Expected failure since cluster creation requires maintainer AWS credentials.

Ready for another look when you get a chance.

python/batch_operations/README.md
```diff
@@ -0,0 +1,140 @@
...
+```bash
+python3 -m venv .venv
+source .venv/bin/activate
+```
+
+Install the required packages:
+
+```bash
+pip install -r requirements.txt
+```
```
amaksimo, line 59: Missing `requirements.txt`.

This README instructs `pip install -r requirements.txt`, but the PR doesn't add one. The sample imports `aurora_dsql_psycopg2` and `psycopg2`, but we cannot install from that dir.

Can you add`requirements.txt` or switch to`pyproject.toml`?

python/batch_operations/src/batch_delete.py
```diff
@@ -0,0 +1,98 @@
...
+Deletes rows matching a condition in configurable-size batches, committing each
+batch as a separate transaction to stay within the 3,000-row mutation limit.
+Uses OCC retry logic with exponential backoff for serialization conflicts.
+
+The parallel variant partitions rows across worker threads using
+``abs(hashtext(id::text)) % num_workers`` so that each worker operates on a
+disjoint subset, avoiding OCC conflicts between workers.
+"""
+
+import threading
```
amaksimo: Unused `threading` isn't referenced in this module (however the`parallel*` modules use it).

python/batch_operations/src/parallel_batch_delete.py
```diff
@@ -0,0 +1,126 @@
...
+        Total number of rows deleted across all workers.
+    """
+    results = [0] * num_workers
+    errors = [None] * num_workers
+
+    def worker(worker_id):
+        total_deleted = 0
+        consecutive_failures = 0
+        partition_condition = (
+            f"{condition} AND abs(hashtext(id::text)) % {num_workers} = {worker_id}"
```
amaksimo: `condition` gets inserted without parentheses before appending the partition filter:

```
... {condition} AND abs(hashtext(...)) % N = i
```

That becomes a problem if someone passes a condition with an OR, like:

```
category = 'food' OR status = 'expired'
```

Because AND has higher precedence than OR, it can cause multiple workers to process the same rows again, defeating the purpose of partitioning and bringing back OCC conflicts.

The current demo is fine since it only uses simple equality conditions, but this is easy for users to run into when adapting the sample. We can do this:

```
partition_condition = (
    f"({condition}) AND abs(hashtext(id::text)) % {num_workers} = {worker_id}"
)
```

Which is better, the same fix should also be applied in:

parallel_batch_update.py
BatchDelete.java
BatchUpdate.java
batchDelete.js
batchUpdate.js

python/batch_operations/src/occ_retry.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    for attempt in range(max_retries + 1):
+        try:
+            return operation(connection)
+        except psycopg2.errors.SerializationFailure:
+            connection.rollback()
+            if attempt >= max_retries:
+                raise MaxRetriesExceeded(max_retries)
+            delay_ms = base_delay_ms * (2 ** attempt)
+            logger.warning(
```
amaksimo, line 57: No jitter in backoff, `delay_ms = base_delay_ms * (2 ** attempt)` is just exp backoff, what about:

```
import random
delay_ms = base_delay_ms * (2 ** attempt) * (0.5 + random.random())
```

the same applies to `OccRetry.java` and `occRetry.js`

python/batch_operations/batch_test_setup.sql
```diff
@@ -0,0 +1,214 @@
...
+-- Create an index on the category column.
+CREATE INDEX ASYNC idx_batch_test_category ON batch_test (category);
+
+-- =============================================================================
+-- Populate the table with 25,000 rows of test data (25 batches of 1,000).
+-- Each INSERT is 1,000 rows — run each as a separate transaction.
+-- =============================================================================
+
+INSERT INTO batch_test (category, status, value)
+SELECT
```
amaksimo: Nit: this 1,000-row INSERT block is repeated 25x and the whole file is duplicated under `java/batch_operations/` and `javascript/batch_operations/` (3 times copied). Options?

1. Replace the 25 copy-pasted INSERTs with a psql `\set` loop, or call from the host language driver (committing each batch as a separate transaction to respect the 3k mut limit).
2. Keep one copy and reference it from each language's README?

javascript/batch_operations/src/occRetry.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Execute a database operation with OCC conflict retry and exponential backoff.
+ *
+ * Retries on SQLSTATE 40001 (serialization failure) with exponential backoff.
+ * Each retry rolls back the current transaction before waiting.
+ *
+ * @param {import('pg').PoolClient} client - A node-postgres pool client.
+ * @param {(client: import('pg').PoolClient) => Promise<*>} operation - Async
+ *   function that performs database work. Should NOT commit.
```
amaksimo: JSDoc says the operation "Should NOT commit," but the operations passed in `batchDelete.js` and `batchUpdate.js` call `BEGIN` inside the operation (and the caller does `COMMIT` after `executeWithRetry` returns). It works — each retry begins a fresh tx after the rollback — but the responsibility split is inconsistent. Either move `BEGIN` into `executeWithRetry`, or update the doc to say "operation must call BEGIN; caller commits after success."

java/batch_operations/src/main/java/com/example/dsql/batch_operations/OccRetry.java
```diff
@@ -0,0 +1,83 @@
...
+import java.sql.SQLException;
+import java.util.logging.Logger;
+
+/**
+ * OCC retry logic with exponential backoff for Aurora DSQL batch operations.
+ */
+public class OccRetry {
+
+    private static final Logger logger = Logger.getLogger(OccRetry.class.getName());
+    private static final String SERIALIZATION_FAILURE = "40001";
```
amaksimo, line 17: This class logs via `java.util.logging.Logger`, but `BatchDelete`, `BatchUpdate`, and `Main` use `System.out.println` everywhere. Pick one?

java/batch_operations/src/test/java/com/example/dsql/batch_operations/MainTest.java
```diff
@@ -0,0 +1,58 @@
...
+        Map<String, String> env = pb.environment();
+        env.put("PGPASSWORD", token);
+        pb.inheritIO();
+        Process psql = pb.start();
+        assertEquals(0, psql.waitFor(), "psql batch_test_setup.sql failed");
+    }
+
+    @Test
+    public void testBatchOperations() {
+        assertAll(() -> Main.main(new String[]{
```
amaksimo: `Main.main` calls `System.exit(1)` on any `SQLException` / `OccRetry.MaxRetriesExceededException`. 

If the integration test hits a db error, JVM exits before JUnit can report it as a failure (exit code 1 with no test output). Take the `main` body into a helper that throws, and have both the CLI entry and the test call it?

anwesham-lab:
Can we upgrade the vulnerable dependencies?

anwesham-lab:
## Fleet review — PR #698 (`Add batch DELETE/UPDATE samples for datasets exceeding 3k row limit`)

Reviewed at head `a59ed68`. **20 reviewers** — a dedicated language deep-dive per language touched (Python, Java, JavaScript) plus cross-cutting lanes (DSQL-semantics, security ×2, cross-language consistency, repo drift, README accuracy, in-code comments, tests, type design, code-simplifier, shallow-bug/porting-drift, silent-failure, prior-comment-addressed, CI/deps, commit-hygiene + build/runnability, independent review) — with the load-bearing DSQL semantics validated against a live cluster (`us-west-2`). Confidence-scored (≥65), with how each was validated.

### Recommendation: **Request changes.** Two issues block a clean first run / merge, and two are real correctness teaching-bugs. The port is otherwise high-fidelity and the code is solid.

The three-language port is unusually faithful (OCC retry, defaults, backoff, parallel partitioning all match), auth/TLS/credential handling is clean, and the gradle wrapper is verified-authentic. But: the **Python sample can't `pip install` as written**, the **npm dep vuln (the CI blocker) is still unfixed**, the **OCC retry doesn't cover commit-time conflicts** (live-proven), and the **`abs(hashtext())` partition can crash a worker** (live-proven).

---

### Blocking / high-value findings

| # | Finding | Lang | File | Conf | How validated |
|---|---------|------|------|------|---------------|
| F1 | **`pip install -r requirements.txt` fails — wrong package name.** `requirements.txt` lists `aurora-dsql-psycopg2>=1.0.0`, which **does not exist on PyPI** (404). The import `aurora_dsql_psycopg2` is provided by `aurora-dsql-python-connector` (what the sibling `python/psycopg2` sample uses). The Python sample cannot be installed as documented. | Py | `python/batch_operations/requirements.txt` | 95 | **Live:** `pypi.org/pypi/aurora-dsql-psycopg2` → HTTP 404; `aurora-dsql-python-connector` → HTTP 200; sibling `python/psycopg2/requirements.txt` confirms the correct name |
| F2 | **CI blocker unaddressed — vulnerable Jest transitive dep.** `package.json` still pins `jest ^30.3.0`; the sibling `javascript/node-postgres` already moved to `^30.4.2`. This is the `deps-review`/`gate` failure that makes the PR BLOCKED, and the exact thing reviewer anwesham-lab asked to fix ("Can we upgrade the vulnerable dependencies?"). Fix: bump to `^30.4.2`, regenerate the lockfile. (Lockfile-commit IS the repo convention — don't drop it.) | JS | `javascript/batch_operations/package.json` | 90 | `gh pr checks` + sibling package.json comparison; prior-comment-addressed lane |
| F3 | **OCC retry does NOT cover commit-time conflicts.** In all three languages the retry wraps only the DML; `commit()` is **outside** `executeWithRetry` (Py `batch_delete.py:62`, Java `BatchDelete.java:47`, JS `batchDelete.js:52`). DSQL raises the serialization conflict at **COMMIT**, not at the DML statement — so a real write-write conflict escapes as a generic error and aborts the batch with **zero retries**, defeating the sample's headline purpose under the very contention it claims to handle. | all | `occ_retry.py` / `OccRetry.java` / `occRetry.js` + callers | 90 | **Live:** forced a write-write conflict → `ERROR: 40001: change conflicts with another transaction (OC000)` fired **at COMMIT** after the UPDATE reported `UPDATE 1` |
| F4 | **`abs(hashtext(id::text))` can crash a parallel worker.** `hashtext` returns int4 and can return `-2147483648`; `abs(INT_MIN)` overflows. That error is not 40001, so it bypasses retry, kills the worker, and aborts the parallel op — deterministically for any id that hashes to INT_MIN. Fix: `abs(hashtext(id::text)::bigint) % n`. | all (parallel) | `parallel_batch_delete.py:48` (+ Java/JS equivalents) | 88 | **Live:** `SELECT abs((-2147483648)::int4)` → `ERROR: integer out of range`; over 100k values `hashtext` already produced `-2147469657` (≈14k from INT_MIN); the `::bigint` fix works |

---

### Correctness / robustness (non-blocking but worth fixing)

| # | Finding | Lang | Conf | How validated |
|---|---------|------|------|---------------|
| F5 | **Incomplete mutation reported as success.** The post-verification `COUNT(*)` only `print`s a WARNING then returns the count and exits 0. On any error path (OCC exhaustion, worker death) it isn't even reached. For a data-mutation sample others copy into migrations, "left rows behind" must fail loudly (raise / non-zero exit), not log-and-continue. | all | 90 | Code trace (silent-failure lane) |
| F6 | **Parallel error handling discards committed work.** When one worker exhausts retries, siblings have already durably committed their partitions; Python raises and throws away the per-worker counts; **JS `Promise.all` rejects but doesn't cancel — workers keep committing after `process.exit(1)`**. A customer sees "it failed" and may blindly re-run, double-applying non-idempotent SET clauses. Use `Promise.allSettled`, collect committed counts, report partial damage. | all (JS worst) | 85 | Code trace |
| F7 | **`SET {set_clause}, updated_at = NOW()`** assumes the target table has an `updated_at` column and that `set_clause` is non-empty/comma-clean; an empty/trailing-comma clause yields malformed SQL. Identical in all three (no drift). | all | 85 | Code trace |
| F8 | **No validation on `--num-workers` / `--batch-size`.** `--num-workers 0` → `% 0` division-by-zero in the partition SQL; negative diverges by language (Py no-ops, JS `RangeError`, Java `IllegalArgumentException`). Unbounded large values size the pool 1:1 → DSQL connection-limit exhaustion. | all | 80 | Code trace (shallow-bug + security lanes) |

---

### Docs / consistency / drift

| # | Finding | Conf | How validated |
|---|---------|------|---------------|
| F9 | **"disjoint partitions avoid OCC conflicts between workers" is overstated** (Python docstrings/READMEs). The file's own `MAX_CONSECUTIVE_FAILURES` retry loop exists precisely to absorb conflicts the comment says are "avoided"; workers still contend on shared index maintenance and DSQL surfaces it as 40001. Reword to "reduces." Also a parity gap: only Python carries the rationale; JS states mechanism-only; Java's parallel methods have no Javadoc. | 85 | DSQL docs + code self-contradiction |
| F10 | **Missing "trusted input only" caveat.** `table`/`condition`/`set_clause` are interpolated unparameterized into SQL in all three (unavoidable for identifiers/clauses), but no code comment or README warns they must be developer-controlled, never end-user input. For widely-copied AWS sample code this teaches SQL injection by example. Add the caveat (and demonstrate binding *values* inside the condition via `%s`/`?`/`$1`). | 88 | Security + comment lanes |
| F11 | **Repo drift.** New files use `MIT-0` SPDX headers while every sibling uses `Apache-2.0`; Java package is the placeholder `com.example.dsql.batch_operations` vs siblings' `software.amazon.dsql.examples`; Java uses Groovy `build.gradle` vs siblings' Kotlin `build.gradle.kts`; Java deps pinned older than siblings; **no CI workflow added** for any of the three (siblings each have a `*-integ-tests.yml`), so these tests never run in the repo pipeline. | 85 | Sibling comparison (drift lane) |
| F12 | **README cleanup block uses `$CLUSTER_REGION`** which isn't re-exported in that section — fails in a fresh shell. Minor; re-export endpoint+region in the Cleanup block. | 70 | README vs code trace |

---

### Verified correct / clean (so they're not re-litigated)
- **OCC SQLSTATE catch is correct** (live-confirmed): DSQL surfaces OC000/OC001 *as* SQLSTATE `40001` — catching `40001` / `SerializationFailure` / `err.code==="40001"` genuinely handles data + schema conflicts. No need to match the OC strings separately. (My initial "they miss OC000/OC001" hypothesis was **falsified** by the live test.)
- **Parallel variant exists in all three languages** (Java/JS fold it into the same file; Python splits it out) — the "Python-only parallel" premise was **wrong**; the port is complete.
- **Auth/credentials/TLS clean**: IAM delegated to official connectors, no token/secret logged, DSQL is server-side TLS-mandatory. **gradle-wrapper.jar verified byte-identical to official Gradle 9.4.0.** `.gitignore` changes and committed wrapper are correct per repo convention.
- **Backoff math, retry counts, defaults, main orchestration, resource cleanup, `batch_test_setup.sql` (byte-identical ×3, DSQL-compatible DDL), self-referencing `DELETE … WHERE id IN (SELECT … LIMIT n)`** — all consistent and correct across the three languages; no porting-drift bug found.
- **The 3,000-row limit is exactly that — table rows, not index entries** (live-confirmed): a `DELETE` of 2,900 rows on a table with **PK + 2 secondary indexes** succeeded, and `DELETE` of **3,001** rows → `ERROR: transaction row limit exceeded` while 3,000 did not. So the README's `batch-size < 3000` guidance and the `batch_size=1000` default are **correct with no index-amplification caveat needed** (an earlier suspected finding to the contrary was falsified on the cluster).
- **Most prior review comments (amaksimo, Benjscho) are genuinely addressed** — condition-parenthesization, jitter, ms rename, logging unification, `System.exit` refactor, repopulate removal, standalone dirs. The two not addressed are F2 (npm vuln) and amaksimo's soft "dedupe the triplicated SQL" suggestion (open question, not a blocker).

---

### Suggested path to merge
1. **Fix `requirements.txt`** → `aurora-dsql-python-connector` (F1) — unblocks the Python sample.
2. **Bump Jest to `^30.4.2`** + regenerate lockfile (F2) — clears the CI gate.
3. **Move `commit()` inside the OCC retry scope** (F3) so commit-time 40001 is retried — this is the sample's core promise.
4. **Cast `hashtext(...)::bigint` before `abs`** (F4).
5. Make the post-verify **fail loudly** (F5), fix parallel partial-failure reporting (F6), validate `--num-workers`/`--batch-size` (F8), and add the trusted-input caveat (F10).
6. Address drift (F11): SPDX header, Java package namespace, add the CI workflows; reword the "avoid OCC conflicts" claim (F9) and the `<3000` guidance (F13).

*Methodology: 20 reviewers — one per language touched (Python/Java/JS) plus 17 cross-cutting lanes spanning the full /security-review, /code-review, and /pr-review-toolkit dimensions (type-design, code-simplifier, comment-analyzer, test-analyzer, silent-failure, code-reviewer), a repo-drift lane, a prior-comment-addressed completeness lane, and an independent generalist. DSQL semantics validated on a live cluster via psql + IAM token — which confirmed F3/F4 and falsified the OCC-code and Python-only-parallel premises. Findings five-gate-validated against head `a59ed68`.*

anwesham-lab (approved):

PR #920, opened 2026-04-17
fix(go): Correct pool config and connector comments in recipe-sharing…

arbeiterz:
…-api

Set MinConns to 0 so the pool creates connections on demand rather than maintaining an idle connection. BeforeConnect generates a fresh IAM token for each new connection, so pre-warming is unnecessary.

Remove inaccurate comment claiming the connector sets prefer_simple_protocol. The connector handles IAM token generation via BeforeConnect, TLS with verify-full, and connection parameter defaults — it does not configure query execution mode.


amaksimo:
Thanks for the contribution! Because this branch is `arbeiterz:main` and has accumulated several unrelated `Sync examples from connectors` merges (which conflict with the current main), I've opened a replacement PR that cherry-picks just your single authored commit onto a clean branch: #1054.

Original authorship (`R. Steven Brown Jr. <rsbjr@amazon.com>`) is preserved on the cherry-picked commit. Suggesting we continue review in #1054 and close this one.

PR #935, opened 2026-04-20
fix: address AppSec security findings for sample-amazon-aurora-dsql-agent

sumeettr-collab:
### Changes

**Least privilege DB access**
- Created `grid_reader` non-admin role with SELECT-only grants in `infra/schema.sql`
- Lambda now connects as `grid_reader` instead of `admin`

**SQL validation hardening**
- Strip SQL comments before validation (prevents hiding forbidden keywords)
- Added PostgreSQL dangerous function blocklist (pg_sleep, dblink, lo_import, etc.)
- Added input length limit (2000 chars)
- Added validation failure logging

**IAM policy scoping**
- Removed `dsql:DbConnectAdmin` from IAM policy and `setup_gateway.py`
- Added scoping guidance comments for production use

**Input sanitization**
- Input length limits on SQL and user questions
- Security Hardening section added to README

**Network binding**
- A2A server defaults to `127.0.0.1`, uses `0.0.0.0` only on AgentCore Runtime

### Testing
- 37 unit tests passing (20 new tests for security hardening)
- Full end-to-end deployment verified against live DSQL cluster

### DIFF
--- a/sample-amazon-aurora-dsql-agent/README.md
+++ b/sample-amazon-aurora-dsql-agent/README.md
@@ -146,7 +146,7 @@ This creates:
 
 | Resource | Name |
 |----------|------|
-| Lambda role | `grid-investigation-lambda-role` (trust: `lambda.amazonaws.com`, policy: `dsql:DbConnectAdmin` + CloudWatch) |
+| Lambda role | `grid-investigation-lambda-role` (trust: `lambda.amazonaws.com`, policy: `dsql:DbConnect` + CloudWatch) |
 | Gateway role | `grid-investigation-gateway-role` (trust: `bedrock-agentcore.amazonaws.com`, policy: `lambda:InvokeFunction`) |
 | Lambda | `grid-investigation-tools` (Python 3.12, 512 MB, 30s timeout) |
 | Gateway | `grid-investigation-gateway` (Cognito OAuth + semantic search enabled) |
@@ -179,6 +179,23 @@ context in its system prompt and generates parameterized SQL dynamically.
 - Multi-statement queries (`;` followed by another statement) are blocked.
 - Results are capped at 500 rows with an auto-appended `LIMIT`.
 - All parameters are passed via psycopg2's parameterized query interface (not string interpolation).
+
+### Security hardening
+
+This sample includes several security measures suitable for production use:
+
+- **Least-privilege database role**: The Lambda connects as `grid_reader` (non-admin) with `dsql:DbConnect` — not `dsql:DbConnectAdmin`. The role only has `SELECT` grants on the 6 allowed tables. This is the primary security boundary.
+- **SQL comment stripping**: Comments (`--` and `/* */`) are stripped before validation to prevent hiding forbidden keywords.
+- **Function blocklist**: Dangerous PostgreSQL functions (`pg_sleep`, `pg_read_file`, `dblink`, etc.) are explicitly blocked.
+- **Input length limits**: SQL queries are capped at 2,000 characters; user questions are capped at 2,000 characters before reaching the LLM.
+- **Validation failure logging**: All SQL validation failures are logged with the rejected query for monitoring and alerting.
+- **Network binding**: The A2A server defaults to `127.0.0.1` for local development (safe by default). AgentCore Runtime passes `--host 0.0.0.0` where network isolation is handled by the platform.
+- **IAM policy scoping**: The sample IAM policy (`infra/iam_policy.json`) uses wildcard resources for convenience. For production, scope each resource to specific ARNs:
+  - DSQL: `arn:aws:dsql:<REGION>:<ACCOUNT_ID>:cluster/<CLUSTER_ID>`
+  - Bedrock: `arn:aws:bedrock:<REGION>::foundation-model/<MODEL_ID>`
+  - Logs: `arn:aws:logs:<REGION>:<ACCOUNT_ID>:log-group:/aws/lambda/<FUNCTION_NAME>:*`
+  - See the Lambda CDK sample for an example of proper resource scoping.
+- **Rate limiting**: For production, add rate limiting at the API Gateway or A2A endpoint layer to limit abuse.
 
 ### Package and update Lambda dependencies
 
@@ -226,9 +243,9 @@ aws lambda invoke \
   /tmp/test_out.json && cat /tmp/test_out.json | python3 -m json.tool
 ```
 
-### (Optional) least-privilege database role
+### Set up the least-privilege database role
 
-By default the Lambda connects as `admin`. For production, create a read-only database role:
+The Lambda connects as `grid_reader`, a non-admin role with SELECT-only grants. This role is defined in `infra/schema.sql` and is created when you apply the schema. You need to grant it to the Lambda's IAM role:
 
 ```bash
 export PGPASSWORD=$(aws dsql generate-db-connect-admin-auth-token \
@@ -239,14 +256,10 @@ psql "host=<cluster-id>.dsql.us-east-1.on.aws port=5432 dbname=postgres user=adm
 ```
 
 ```sql
-CREATE ROLE grid_app WITH LOGIN;
-AWS IAM GRANT grid_app TO 'arn:aws:iam::<account-id>:role/grid-investigation-lambda-role';
-GRANT USAGE ON SCHEMA public TO grid_app;
-GRANT SELECT ON ALL TABLES IN SCHEMA public TO grid_app;
+AWS IAM GRANT grid_reader TO 'arn:aws:iam::<account-id>:role/grid-investigation-lambda-role';
 ```
 
-Then update `handler.py` to use `user="grid_app"` and change the IAM policy from
-`dsql:DbConnectAdmin` to `dsql:DbConnect`.
+This is the primary security boundary — even if SQL validation is bypassed, the database enforces SELECT-only access on the allowed tables.
 
 ---
 
@@ -286,6 +299,9 @@ For local development and testing, you can run the Database Agent directly inste
 ```bash
 python agent/database_agent.py [--region us-east-1] [--config gateway/gateway_config.json]
 ```
+
+The server binds to `127.0.0.1` by default (local-only access). When deployed to
+AgentCore Runtime, the platform passes `--host 0.0.0.0` to enable container routing.
 
 The Database Agent will:
 1. Connect to the AgentCore Gateway via MCP (using Cognito OAuth from `gateway_config.json`)
@@ -350,6 +366,11 @@ To avoid ongoing charges, delete the resources created in this walkthrough:
 ```bash
 # Delete the AgentCore agent (also removes memory, execution role, and S3 artifacts)
 agentcore destroy -a grid_database_agent
+
+# IMPORTANT: Remove the cached agentcore config — it stores the old agent ID.
+# If you skip this and redeploy, `agentcore launch` will try to update the
+# deleted agent and fail with ResourceNotFoundException.
+rm -f .bedrock_agentcore.yaml
 
 # Delete the Lambda function
 aws lambda delete-function --function-name grid-investigation-tools --region us-east-1
--- a/sample-amazon-aurora-dsql-agent/agent/database_agent.py
+++ b/sample-amazon-aurora-dsql-agent/agent/database_agent.py
@@ -40,8 +40,13 @@ logging.basicConfig(
 MODEL_ID = "global.anthropic.claude-sonnet-4-6"
 
 # A2A server defaults
+# Binds to 0.0.0.0 by default so AgentCore Runtime can route traffic to the
+# container. For local development, use --host 127.0.0.1 to restrict access.
 A2A_HOST = "0.0.0.0"
 A2A_PORT = 9000
+
+# Maximum length for user questions to limit prompt injection surface
+MAX_QUESTION_LENGTH = 2000
 
 
 SYSTEM_PROMPT_TEMPLATE = """You are a grid operations investigation assistant with access to an Aurora DSQL
@@ -211,7 +216,7 @@ def create_a2a_app(config: dict, region: str) -> FastAPI:
     2. Fetches live schema via get_schema MCP tool
     3. Creates the Strands Agent with dynamic schema
     4. Wraps it in an A2AServer for A2A protocol support
-    5. Mounts on a FastAPI app with health check
+    5. Mounts on a FastAPI app with health check and input length guard
 
     Returns:
         FastAPI app ready to serve A2A requests.
@@ -260,14 +265,46 @@ def create_a2a_app(config: dict, region: str) -> FastAPI:
             serve_at_root=True,
         )
 
-        # Step 5: Mount on FastAPI
+        # Step 5: Mount on FastAPI with ASGI-level input length guard
         app = FastAPI()
 
         @app.get("/ping")
         def ping():
             return {"status": "healthy"}
 
-        app.mount("/", a2a_server.to_fastapi_app())
+        # Wrap the A2A app in an ASGI middleware class that rejects oversized
+        # requests via Content-Length without reading the body. This avoids the
+        # known Starlette BaseHTTPMiddleware issue where reading the body drains
+        # the ASGI receive stream and breaks downstream mounted apps.
+        inner_a2a_app = a2a_server.to_fastapi_app()
+
+        class ContentLengthGuard:
+            """ASGI middleware that rejects requests exceeding a size limit."""
+
+            def __init__(self, app):
+                self.app = app
+
+            async def __call__(self, scope, receive, send):
+                if scope["type"] == "http" and scope.get("method") == "POST":
+                    headers = dict(scope.get("headers", []))
+                    content_length = headers.get(b"content-length", b"0")
+                    try:
+                        if int(content_length) > MAX_QUESTION_LENGTH * 2:
+                            from starlette.responses import JSONResponse
+                            response = JSONResponse(
+                                status_code=413,
+                                content={
+                                    "error": f"Request body too large. "
+                                    f"Maximum question length is {MAX_QUESTION_LENGTH} characters."
+                                },
+                            )
+                            await response(scope, receive, send)
+                            return
+                    except (ValueError, TypeError):
+                        pass  # Missing or malformed Content-Length — let it through
+                await self.app(scope, receive, send)
+
+        app.mount("/", ContentLengthGuard(inner_a2a_app))
     except Exception:
         mcp_client.__exit__(None, None, None)
         raise
@@ -300,6 +337,11 @@ def parse_args() -> argparse.Namespace:
         default=A2A_PORT,
         help=f"Port to serve on (default: {A2A_PORT})",
     )
+    parser.add_argument(
+        "--host",
+        default="127.0.0.1",
+        help="Host to bind to (default: 127.0.0.1). AgentCore Runtime passes --host 0.0.0.0.",
+    )
     return parser.parse_args()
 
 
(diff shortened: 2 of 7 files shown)


anwesham-lab (dismissed):
commit message should not include ticket IDs in header or body, simply title as addressing security findings.

anwesham-lab:
### Code review

Ran 12 parallel review agents (CLAUDE.md adherence, bugs, silent failures, types, tests, comment accuracy, schema/DSQL, IAM, git history, docs claims, A2A binding, SQL validator fuzzing). No CLAUDE.md exists in this repo. The overall direction — least-privilege DB role + defense-in-depth validation — is solid.

> **Note on confidence:** these findings come from automated review agents and were not all executed/verified end-to-end. Each item below is tagged with a confidence level. Please treat `Needs verification` items as leads to confirm, not as certainties. I may also have missed context you already considered.

---

#### 1. `MAX_QUESTION_LENGTH` declared but never enforced — README overstates behavior  *(Confidence: High)*

The constant is defined, but no code path truncates or rejects user questions against it. The README [Security hardening](https://github.com/aws-samples/aurora-dsql-samples/blob/8b1df14694ae0058812c320ef7f65c2ddc42126b/sample-amazon-aurora-dsql-agent/README.md#L183-L193) section claims "user questions are capped at 2,000 characters before reaching the LLM."

https://github.com/aws-samples/aurora-dsql-samples/blob/8b1df14694ae0058812c320ef7f65c2ddc42126b/sample-amazon-aurora-dsql-agent/agent/database_agent.py#L48-L51

Fix: either enforce it in the A2A request handler / FastAPI middleware, or remove the constant and README bullet so docs match behavior. (Possible you meant to wire this in a follow-up PR — flagging if not.)

#### 2. `create_a2a_app` return type annotation is stale  *(Confidence: High)*

Function now returns `(app, bind_host)`, docstring was updated, but the signature annotation wasn't.

https://github.com/aws-samples/aurora-dsql-samples/blob/8b1df14694ae0058812c320ef7f65c2ddc42126b/sample-amazon-aurora-dsql-agent/agent/database_agent.py#L212-L225

Change `-> FastAPI:` → `-> tuple[FastAPI, str]:`.

#### 3. `AGENTCORE_RUNTIME_URL` may not be set inside the container — deployed agent could bind 127.0.0.1 and fail  *(Confidence: Needs verification — this is the one most worth double-checking before merge)*

The new host-detection logic gates on `AGENTCORE_RUNTIME_URL` being present in the container's environment:

https://github.com/aws-samples/aurora-dsql-samples/blob/8b1df14694ae0058812c320ef7f65c2ddc42126b/sample-amazon-aurora-dsql-agent/agent/database_agent.py#L263-L267

I could not find authoritative documentation that AgentCore Runtime injects `AGENTCORE_RUNTIME_URL` *into* the container. A couple of secondary review signals (agent findings that searched docs; the fact that this same variable is used client-side at [L258-L260](https://github.com/aws-samples/aurora-dsql-samples/blob/8b1df14694ae0058812c320ef7f65c2ddc42126b/sample-amazon-aurora-dsql-agent/agent/database_agent.py#L256-L261) as an invocation URL — a fairly strong hint it's a user-set client-side value) suggest the A2A contract requires containers to bind `0.0.0.0:9000` unconditionally. If `AGENTCORE_RUNTIME_URL` is never set in the running container, `is_deployed` is always `False` → binds `127.0.0.1` → platform healthchecks fail silently.

If you've verified empirically that the deployed container receives this env var, ignore this finding. If not, the safest fix is to bind `0.0.0.0` unconditionally (matches official A2A examples) or gate on a confirmed platform variable / explicit `A2A_BIND_HOST` in the Dockerfile. You note in the PR description that end-to-end deployment was verified — if that succeeded, please confirm the env var is actually set at runtime; otherwise the conditional may simply be dead and the platform is tolerating it some other way.

#### 4. SQL validator bypass: forbidden-function name split by a block comment  *(Confidence: Medium-High — logic is straightforward but I did not run the payload against a real DSQL cluster)*

`_strip_sql_comments` replaces `/**/` with a single space, so a payload like `SELECT pg/**/_sleep(5) FROM grid_incidents` becomes `SELECT pg _sleep(5) FROM grid_incidents`, which no longer matches `FORBIDDEN_FUNCTION_PATTERN` (`\bpg_sleep\s*\(`). PostgreSQL treats `/**/` the same way and parses the token as `pg_sleep`.

https://github.com/aws-samples/aurora-dsql-samples/blob/8b1df14694ae0058812c320ef7f65c2ddc42126b/sample-amazon-aurora-dsql-agent/lambda/grid_tools/handler.py#L101-L110

https://github.com/aws-samples/aurora-dsql-samples/blob/8b1df14694ae0058812c320ef7f65c2ddc42126b/sample-amazon-aurora-dsql-agent/lambda/grid_tools/handler.py#L55-L65

Under `grid_reader`, `pg_sleep` / `pg_read_file` / `current_setting` are executable by default — DoS / info-disclosure primitive. The PR already acknowledges the DB role is the "primary security boundary," so severity is bounded, but this class of bypass is still worth closing. Options: (a) strip comments iteratively before the function regex, (b) also scan the raw input, (c) `REVOKE EXECUTE` on the sensitive `pg_catalog` functions from `grid_reader` in `schema.sql` as DB-level defense. A quick unit test would confirm whether this is real — worth running before discarding.

Also: the stripper's docstring says "handles nested poorly but sufficient for validation" — the non-greedy regex doesn't handle nesting at all, so the wording is misleading regardless of the bypass above.

#### 5. Validation-failure log truncates rejected SQL at 200 chars — forensic blind spot  *(Confidence: High)*

`MAX_SQL_LENGTH` is 2000 but the warning truncates to 200 with no ellipsis/marker.

https://github.com/aws-samples/aurora-dsql-samples/blob/8b1df14694ae0058812c320ef7f65c2ddc42126b/sample-amazon-aurora-dsql-agent/lambda/grid_tools/handler.py#L196-L202

Log the full rejected SQL (already capped at 2000) or mark truncation explicitly: `sql=%s…[truncated len=%d]`.

#### 6. `iam_policy.json` top-level `"_comment"` key will be rejected by `aws iam create-policy`  *(Confidence: Medium — depends on how readers actually use the file)*

IAM policy grammar ([docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_grammar.html)) accepts only `Version`, `Id`, `Statement` at the top level. `aws iam create-policy --policy-document file://iam_policy.json` would return `MalformedPolicyDocument`.

https://github.com/aws-samples/aurora-dsql-samples/blob/8b1df14694ae0058812c320ef7f65c2ddc42126b/sample-amazon-aurora-dsql-agent/infra/iam_policy.json#L1-L10

The file isn't consumed programmatically in the deploy path (live policy is built in `setup_gateway.py`), so this doesn't break the PR itself — it only affects readers who copy-paste it. Could be intentional (it's a reference policy, not a deploy artifact). If so, a note in the file header or renaming to `iam_policy.sample.json` would preempt confusion.

---

#### Minor notes (low confidence / stylistic)

- `schema.sql`: `CREATE ROLE grid_reader WITH LOGIN` isn't idempotent — second run fails. PostgreSQL/DSQL don't support `CREATE ROLE IF NOT EXISTS`. Wrap in a `DO ... EXCEPTION WHEN duplicate_object` block if rerunnability matters. May be intentional for a one-shot setup script. ([schema.sql L80-L99](https://github.com/aws-samples/aurora-dsql-samples/blob/8b1df14694ae0058812c320ef7f65c2ddc42126b/sample-amazon-aurora-dsql-agent/infra/schema.sql#L80-L99))
- `handler.py` docstring lists "Parameterized queries via psycopg2" as defense-in-depth layer #8, but `validate_sql` itself doesn't parameterize — that happens in `query_grid_database`. Minor scoping confusion. ([handler.py L113-L126](https://github.com/aws-samples/aurora-dsql-samples/blob/8b1df14694ae0058812c320ef7f65c2ddc42126b/sample-amazon-aurora-dsql-agent/lambda/grid_tools/handler.py#L113-L126))
- Bypass candidate to consider (not verified): `SELECT "pg_sleep"(5) FROM grid_incidents` — quoted-identifier form of a forbidden function. Whether the current regex catches it depends on word-boundary semantics around `"`; worth a one-line test either way.

#### Positive observations

- `dsql:DbConnectAdmin` → `dsql:DbConnect` is coherent end-to-end with the `grid_reader` role.
- 20 new unit tests meaningfully cover happy paths for each new control; `test_cte_alias_rejected_known_limitation` is a great explicit regression doc.
- Per-table GRANTs are the correct DSQL pattern (DSQL does not support `ALTER DEFAULT PRIVILEGES`).
- Loopback-by-default bind is the right direction; just needs deployment detection verified (#3).

---

Disclaimer: these findings are from parallel automated reviewers — some reasoning about runtime behavior (AgentCore env vars, exact Postgres parsing semantics) should be confirmed empirically before acting. If any finding conflicts with context not visible in the diff, please push back.

🤖 Generated with [Claude Code](https://claude.ai/code) — PR-935 @ [`8b1df14`](https://github.com/aws-samples/aurora-dsql-samples/commit/8b1df14694ae0058812c320ef7f65c2ddc42126b)

<sub>If this review was useful, please react with 👍. Otherwise, 👎.</sub>

anwesham-lab:
### Code review (follow-up on [`b1f7307`](https://github.com/aws-samples/aurora-dsql-samples/commit/b1f7307a19b3b878497b60933c9cdd9ab9bd4028))

Thanks for the fast turnaround — all 6 prior findings are addressed cleanly: `MAX_QUESTION_LENGTH` is now enforced, `create_a2a_app` signature updated, `AGENTCORE_RUNTIME_URL` detection removed, `pg/**/_sleep(5)` bypass closed via pre-strip scan + empty replacement, full SQL logged on validation failure, and `_comment` removed from `iam_policy.json`. Two new targeted tests landed as well. Nice.

Re-ran 8 agents on the new head. Flagging 2 critical regressions introduced by the fixes, plus a few smaller items. As before, confidence is tagged per finding — please push back on anything that reflects missing context.

---

#### 1. FastAPI middleware drains the ASGI receive stream — breaks every real A2A POST in production  *(Confidence: High)*

```python
@app.middleware("http")
async def limit_question_length(request, call_next):
    if request.method == "POST":
        body = await request.body()       # ← drains ASGI receive
        ...
    return await call_next(request)
```

https://github.com/aws-samples/aurora-dsql-samples/blob/b1f7307a19b3b878497b60933c9cdd9ab9bd4028/sample-amazon-aurora-dsql-agent/agent/database_agent.py#L276-L294

Starlette's `BaseHTTPMiddleware` (which backs `@app.middleware("http")`) caches the body on the outer `Request` object, but does **not** re-wrap the downstream `receive` callable. The A2AServer mounted at `/` (via `app.mount("/", a2a_server.to_fastapi_app())`) constructs its own `Request`, calls `receive()` on an already-drained channel, and sees an empty body — so every real A2A JSON-RPC call will parse-error (-32700) regardless of length.

This is a [known long-standing Starlette behavior](https://github.com/encode/starlette/issues/495). Fix options, in order of preference:

1. **Do the length check inside the agent/task handler** where the parsed A2A `Message` is already available (simplest, safest).
2. Implement the middleware at pure ASGI level: inspect `Content-Length` (+ fast-reject) without calling `.body()`, or buffer the body and replay by wrapping `scope["receive"]`.
3. If you must read the body here, fall back to a custom ASGI middleware (not `@app.middleware("http")`) that rebuilds a receive function from the buffered bytes and passes it down.

Worth verifying locally before merging — sending a single A2A POST at the deployed runtime will surface the regression immediately.

#### 2. `FORBIDDEN_FUNCTION_PATTERN` regex change introduces false positives on legitimate identifiers  *(Confidence: High)*

The pattern went from `\b(name)\s*\(` to `"?(name)"?\s*\(`:

https://github.com/aws-samples/aurora-dsql-samples/blob/b1f7307a19b3b878497b60933c9cdd9ab9bd4028/sample-amazon-aurora-dsql-agent/lambda/grid_tools/handler.py#L55-L65

Removing `\b` on the left makes the regex match the forbidden name as a *suffix* of any longer identifier, so long as `(` follows directly. Concrete false positives:

- `SELECT my_current_setting(x) FROM grid_incidents LIMIT 1` — at position of `c` in `current_setting` inside `my_current_setting`, `"?` matches empty, `current_setting` matches, `"?` empty, `\s*` empty, `(` matches. Rejected.
- `SELECT app_dblink(x) FROM grid_incidents LIMIT 1` — same shape, rejected.
- Similarly any user-defined function whose name ends with `pg_sleep`, `lo_import`, `set_config`, `current_setting`, etc.

(Identifiers that don't end with the forbidden name — e.g. `my_pg_sleep_counter(x)`, `set_config_value(x)` — are correctly NOT matched, because the char after the forbidden substring is `_`, not `(`.)

No negative test asserts these pass. Suggested fix: restore left-side anchoring while allowing the optional quote, e.g.

```python
r"(?<![\w\"])\"?(" + "|".join(re.escape(f) for f in FORBIDDEN_FUNCTIONS) + r")\"?\s*\("
```

(lookbehind `(?<![\w"])` keeps both the bareword and quoted-identifier cases working without swallowing suffixes). Then add negative tests covering `my_current_setting(x)` and `app_dblink(x)`.

---

#### 3. Local dev now binds `0.0.0.0` by default — reverts part of the original hardening  *(Confidence: Medium-High)*

`A2A_HOST` changed from `127.0.0.1` → `0.0.0.0`, and the new `--host` CLI defaults to `A2A_HOST`:

https://github.com/aws-samples/aurora-dsql-samples/blob/b1f7307a19b3b878497b60933c9cdd9ab9bd4028/sample-amazon-aurora-dsql-agent/agent/database_agent.py#L42-L46

https://github.com/aws-samples/aurora-dsql-samples/blob/b1f7307a19b3b878497b60933c9cdd9ab9bd4028/sample-amazon-aurora-dsql-agent/agent/database_agent.py#L329-L333

A developer running `python agent/database_agent.py` locally will now bind to all interfaces, exposing the agent (which holds DSQL access via `grid_reader`) on LAN/Wi-Fi. This is a functional regression of the local-safe default that the first iteration of this PR added.

Suggested fix: keep `A2A_HOST = "0.0.0.0"` as the module-level "what the runtime expects" constant, but set `parser.add_argument("--host", default="127.0.0.1", ...)` so local runs are safe by default; have the AgentCore container entrypoint / Dockerfile explicitly pass `--host 0.0.0.0`.

The README's new "For local development, use `--host 127.0.0.1` to restrict access" guidance is accurate documentation, but safe-by-default matters more than documentation for security-relevant wiring.

---

#### 4. Middleware's `except (json.JSONDecodeError, AttributeError): pass` is overly broad and silent  *(Confidence: Medium — assuming #1 is fixed and the middleware actually runs)*

https://github.com/aws-samples/aurora-dsql-samples/blob/b1f7307a19b3b878497b60933c9cdd9ab9bd4028/sample-amazon-aurora-dsql-agent/agent/database_agent.py#L280-L294

Catching `AttributeError` masks any shape variance in the JSON payload (list instead of dict, `None` in an unexpected slot) as well as real programmer errors, and the bare `pass` means the length guard silently becomes a no-op. Recommend:

- Gate on `Content-Type: application/json` instead of relying on `json.loads` success.
- Use `isinstance` checks on `parts` / each `part` rather than `AttributeError` as a catch-all.
- If JSON parsing fails on a POST-JSON request, log and reject (don't silently pass through).

#### 5. `bind_host` returned from `create_a2a_app` is dead — caller uses `args.host`  *(Confidence: High — low-impact simplification)*

https://github.com/aws-samples/aurora-dsql-samples/blob/b1f7307a19b3b878497b60933c9cdd9ab9bd4028/sample-amazon-aurora-dsql-agent/agent/database_agent.py#L350-L359

`main()` destructures `app, bind_host = create_a2a_app(...)` but then passes `args.host` (not `bind_host`) to `uvicorn.run`. The returned string is only used to log and is identical to `A2A_HOST` regardless of what `--host` was set to. Either:

- Revert `create_a2a_app` to `-> FastAPI:` and drop the tuple (simplest), or
- Actually own the bind decision inside `create_a2a_app` by passing `args.host` in and returning the resolved value (then use `bind_host`, not `args.host`, in the uvicorn call).

---

#### Minor / non-blocking

- `from starlette.responses import JSONResponse` is inlined inside the middleware function body — hoist to module-level imports (prefer `fastapi.responses.JSONResponse` since FastAPI already pulls Starlette in).
- Quoted-identifier test uses bare `pytest.raises(ValueError)` while its siblings use `match="Forbidden function"` — adding the `match=` locks in which layer of the validator catches it and guards against silent drift if the allowlist/strip path happens to catch it instead.
- Consider adding a test for the A2A middleware itself (happy path + at-limit + over-limit + non-JSON POST). This would likely have caught finding #1 before it shipped.
- Middleware returns HTTP 400 for oversize input; HTTP 413 (`Payload Too Large`) is the more semantically correct status.

#### Positive observations

- `schema.sql` now wraps `CREATE ROLE grid_reader` in a `DO $$ ... EXCEPTION WHEN duplicate_object` block — clean idempotency fix.
- Comment replacement changed from `" "` to `""`, with the rationale inlined in the docstring. Well-documented.
- Double-pass (raw + stripped) check on forbidden functions is a good defense-in-depth pattern — worth keeping even after fixing finding #2.
- Two new unit tests for the concrete bypasses flagged previously.
- IAM policy is now strict-valid JSON; Bedrock + CloudWatch statements added make the sample policy more realistic.

---

Disclaimer: these findings are from parallel automated reviewers — particularly #1 (Starlette middleware body-drain) and #2 (regex false-positive) are worth reproducing locally before acting. A 10-line pytest for #2 and a single manual A2A POST for #1 would both be decisive. If any of this conflicts with runtime context you've verified, please push back.

🤖 Generated with [Claude Code](https://claude.ai/code) — PR-935 @ [`b1f7307`](https://github.com/aws-samples/aurora-dsql-samples/commit/b1f7307a19b3b878497b60933c9cdd9ab9bd4028)

<sub>If this review was useful, please react with 👍. Otherwise, 👎.</sub>

amaksimo:
Few comments for things we want to fix:

1. DO $$ block in schema.sql — will fail entirely (confirmed, worse)
DSQL doesn't support PL/pgSQL at all. The DO $$ block won't just fail on second run — it will fail on first run too. The sequelize sample in the same repo explicitly documents this: "PL/pgSQL is not supported." This is a hard blocker for the schema setup.

2. ASGI middleware monkeypatch — dead code (confirmed)
Verified with a live Python test: app() returns "original", app.__call__() returns "patched". Python resolves __call__ via type(), not the instance dict. The 413 guard never fires.

3. Comment stripping vs string literals — bypass works (confirmed, mitigated)
Verified: SELECT * FROM grid_incidents WHERE x = '/*' UNION SELECT * FROM pg_tables WHERE y = '*/' strips to SELECT * FROM grid_incidents WHERE x = '' LIMIT 10. The table allowlist and DML/DDL checks are fully bypassed. Mitigated by the grid_reader role, but the validator is providing a false sense of security. Worth documenting as a known limitation at minimum.

amaksimo:
> Few comments for things we want to fix:
> 
> 1. DO $$ block in schema.sql — will fail entirely (confirmed, worse)
>    DSQL doesn't support PL/pgSQL at all. The DO $$ block won't just fail on second run — it will fail on first run too. The sequelize sample in the same repo explicitly documents this: "PL/pgSQL is not supported." This is a hard blocker for the schema setup.
> 2. ASGI middleware monkeypatch — dead code (confirmed)
>    Verified with a live Python test: app() returns "original", app.**call**() returns "patched". Python resolves **call** via type(), not the instance dict. The 413 guard never fires.
> 3. Comment stripping vs string literals — bypass works (confirmed, mitigated)
>    Verified: SELECT * FROM grid_incidents WHERE x = '/_' UNION SELECT * FROM pg_tables WHERE y = '_/' strips to SELECT * FROM grid_incidents WHERE x = '' LIMIT 10. The table allowlist and DML/DDL checks are fully bypassed. Mitigated by the grid_reader role, but the validator is providing a false sense of security. Worth documenting as a known limitation at minimum.

LGTM now, all are addressed

anwesham-lab (approved):

PR #1031, opened 2026-05-01
feat(deno): Add Deno + Aurora DSQL samples with booking API

gxjx-x:
## Summary

Adds Deno runtime samples for Amazon Aurora DSQL — a basic client example and a full booking-API sample application. Both samples use the official [`@aws/aurora-dsql-postgresjs-connector`](https://github.com/awslabs/aurora-dsql-connectors/tree/main/node/postgres-js), which handles IAM token generation and refresh, SSL/TLS, pooling, and region auto-discovery.

### `deno/postgres-js/` — Basic client sample

Connect-and-query example using `@aws/aurora-dsql-postgresjs-connector` over IAM-authenticated TLS. Demonstrates concurrent queries across a pooled client. Follows the layout used by `javascript/postgres-js/`.

- `src/example_preferred.ts` — main example
- `test/example_preferred.test.ts` — integration test
- `deno.json`, `deno.lock`, `README.md`, `LICENSE`

### `deno/booking-api/postgres-js/` — Booking API sample

A booking/reservation REST API built with `Deno.serve()` and the postgres.js connector, backed by Aurora DSQL. Beyond connect-and-query, this sample shows real application patterns:

- Full CRUD over `/bookings` plus a `GET /health` endpoint
- Layered defense against double-booking: in-transaction overlap check, a `UNIQUE` async index on `(resource_name, start_time, end_time)` as a commit-time backstop, and automatic retry on OCC conflicts (SQLSTATE `40001`)
- Pooled IAM-authenticated connections; the connector refreshes tokens transparently
- Request hardening: 16 KB JSON body cap, strict UUID v4 path regex, redacted error logging
- Safe-by-default bind to `127.0.0.1` (override via `HOST`) with a production note on reverse-proxy deployment
- Graceful HTTP shutdown: drains in-flight requests before closing the pool on `SIGINT`/`SIGTERM`
- UUID primary keys via `gen_random_uuid()` — see the [Aurora DSQL guide on sequences and identity columns](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/sequences-identity-columns.html) if you prefer compact integer keys
- Application-layer referential integrity, as recommended by the [Aurora DSQL migration guide](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-migration-guide.html)

### `deno/CONTRIBUTORS.md`

Credits original and follow-up authors.

### CI workflows

Adds `deno-postgresjs-integ-tests.yml` and `deno-booking-api-integ-tests.yml` mirroring the existing `javascript-postgresjs` pattern. Both workflows run Deno integration tests against a fresh DSQL cluster (created and torn down per run), and both are wired into `ci-gate.yml`. For now they piggy-back on `secrets.JAVASCRIPT_IAM_ROLE` rather than introducing a separate `DENO_IAM_ROLE` — a maintainer can move them to a dedicated role in a follow-up if needed.

## Testing

All tests verified against a live Aurora DSQL cluster:

- **18 unit + property tests + 13 integration tests** (`fast-check`, `@std/assert`): router dispatch, overlap symmetry, JSON response format, malformed-input rejection, OCC retry/exhaustion/pass-through, concurrent overlap-race behavior, scoped test cleanup
- **14 smoke tests** (`test-api.sh`): health check, CRUD lifecycle, overlap conflict, concurrent OCC, validation errors

## Documentation

The booking-API README covers architecture, request flow, data model, Aurora DSQL design decisions, API endpoints with status codes, error response shapes, the Deno permissions model, OCC retry behavior, optional Deno Deploy deployment, and example `curl` commands. Production guidance calls out keyset pagination for `GET /bookings` and running CRUD as a non-admin database role — see the [Aurora DSQL guide on database roles and IAM authentication](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/using-database-and-iam-roles.html).

## Notes for reviewers

- The repository root `README.md` is updated to list the two Deno samples alongside existing entries in the Client/ORM and Sample Applications tables
- Authentication and cluster-management samples are intentionally omitted — these would be AWS SDK calls with no Deno-specific value, matching the conventions already used for TypeScript and PHP
- Commit history is intentionally granular (one commit per theme) to ease review

## Credits

Builds on [#950](https://github.com/aws-samples/aurora-dsql-samples/pull/950) by [@pyraenix](https://github.com/pyraenix), credited as co-author on every commit.

Some commits in this PR were authored with assistance from [Kiro](https://kiro.dev).

---


deno/booking-api/postgres-js/README.md
```diff
@@ -0,0 +1,478 @@
+# Booking API
+
+## Deno + postgres.js connector + Amazon Aurora DSQL
+
+A booking/reservation REST API built with **Deno.serve()** and the
+[**Aurora DSQL Connector for postgres.js**](https://github.com/awslabs/aurora-dsql-connectors/tree/main/node/postgres-js),
+backed by **Amazon Aurora DSQL**. This sample demonstrates IAM token
```
amaksimo: Why do we need to say backed by?

deno/booking-api/postgres-js/README.md
```diff
@@ -0,0 +1,478 @@
...
+  │  201 {booking}      │     │                                │
+  │◀──────────────────────────│                                │
+  │                     │     │                                │
+  │                     │ No  │  Overlap found                 │
+  │                     └─────┤  (auto ROLLBACK on return)     │
+  │  409 {conflict}           │                                │
+  │◀──────────────────────────│                                │
+  │                           │                                │
+  │                     ┌─────────────────────────────┐        │
+  │                     │  If DSQL returns OC000/     │        │
```
amaksimo: We should use "40001" as well

deno/booking-api/postgres-js/README.md
```diff
@@ -0,0 +1,478 @@
...
+`SELECT ... FOR UPDATE` on a parent `resources` row) or route writes for
+the same resource through a queue.
+
+The integration test `occ-overlap-race.integration.test.ts` documents
+both cases: the success case (identical windows → exactly one 201,
+rest 409/503) and the overlapping-but-distinct case.
+
+---
+
+## Key Design Decisions for Aurora DSQL
```
amaksimo, line 201: Not sure if we want to keep this section. Imagine we add features in the future, we may need to maintain this. I think we can just say that the example uses the connector which we said above and this does not add value

deno/booking-api/postgres-js/README.md
```diff
@@ -0,0 +1,478 @@
...
+# Required
+export CLUSTER_ENDPOINT="<your-cluster>.dsql.us-east-1.on.aws"
+export CLUSTER_USER="admin"
+
+# Optional
+export PORT=8000                  # HTTP server port (default: 8000)
+export HOST="127.0.0.1"           # HTTP bind address (default: localhost)
+                                  # Set "0.0.0.0" only when deploying behind
+                                  # a trusted reverse proxy
+export AWS_REGION="us-east-1"     # Auto-discovered from CLUSTER_ENDPOINT
```
amaksimo: optional then?

deno/booking-api/postgres-js/README.md
```diff
@@ -0,0 +1,478 @@
...
+## OCC Retry Behavior
+
+Aurora DSQL uses optimistic concurrency control (OCC) for transaction
+isolation. When two transactions conflict — for example, two users
+updating the same booking at the same time — Aurora DSQL aborts one
+with SQLSTATE `OC000` (row-level conflict) or `OC001` (schema namespace
+conflict). Both share class `40001` (serialization failure) and are safe
+to retry. Applications retry the aborted transaction.
+
+This sample handles those errors automatically via the `withOccRetry`
```
amaksimo: We want to add match for "40001" within that function + we can probably replace as follow-up once we have OCC retry helper in postgres (maybe add a task in asana to not forget this).

deno/booking-api/postgres-js/README.md
```diff
@@ -0,0 +1,478 @@
...
+utility:
+
+* Detects OCC SQLSTATEs at `error.code` (postgres.js attaches the
+  SQLSTATE directly, not in a nested field)
+* Retries the failed transaction up to 3 times (configurable)
+* Applies exponential backoff with jitter to reduce contention
+* Logs each retry attempt with attempt number and elapsed time
+* Throws `OCC retry exhausted` if all retries fail — mapped to HTTP 503
```
amaksimo: why add a custom error type to just wrap, instead we can just propagate the last occ error we got, I think this is much cleaner.

deno/booking-api/postgres-js/README.md
```diff
@@ -0,0 +1,478 @@
...
+├── test/
+│   └── integration.test.ts          # CRUD integration tests (cluster)
+├── occ-overlap-race.integration.test.ts  # Concurrency tests (cluster)
+├── *.property.test.ts               # Property-based tests (no cluster)
+└── README.md
+```
+
+---
+
+## OCC Retry Behavior
```
amaksimo, line 399: I think we can cut the size of this section and just mention that it includes feature to retry occ and maybe link to occ doc (aws docs).

deno/booking-api/postgres-js/occ-retry.ts
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @module occ-retry
+ */
+
+/**
+ * Configuration options for the OCC retry handler.
+ *
+ * @property maxRetries - Maximum retry attempts after the initial try.
+ *   `maxRetries: 3` means up to 4 total attempts. Default: 3.
+ * @property baseDelayMs - Base delay for exponential backoff. Default: 100.
```
amaksimo: I think we can start with 0 or close to 0 initially, then each increment is just a small bump above that (that is that cpcl do).

Aside, we should replace this entire code with code from connector once Prabakaran adds it.

deno/booking-api/postgres-js/occ-retry.ts
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * Both share SQLSTATE class 40001 (serialization_failure); both are safe
+ * to retry because the conflicting transaction has already aborted.
+ *
+ * postgres.js puts the SQLSTATE code at `error.code` directly.
+ */
+export function isOccError(error: unknown): boolean {
+  if (!error || typeof error !== "object") return false;
+  const code = (error as Record<string, unknown>).code;
+  return code === "OC000" || code === "OC001";
```
amaksimo: Add 40001 to align with connectors.

deno/booking-api/postgres-js/README.md
```diff
@@ -0,0 +1,478 @@
...
+| **Application-layer referential integrity** | For referential integrity patterns, Aurora DSQL's [migration guide](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-migration-guide.html) recommends enforcing relationships in the application layer. This sample stores `booked_by` as a plain string rather than a foreign key; a production app would validate against a `users` table in the same transaction |
+| **Application-layer overlap detection** | To express "no two rows may have overlapping `[start, end)` ranges for the same resource," the sample combines an application-layer SELECT, a unique-window index, and OCC retry. See the concurrency model section for the full pattern |
+| **`CREATE [UNIQUE] INDEX ASYNC`** | Aurora DSQL uses `CREATE INDEX ASYNC` for non-blocking index creation. `setupSchema` waits for the returned `job_id` to complete before handling traffic — see the [Asynchronous indexes](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-create-index-async.html) guide |
```
amaksimo: Exhaustive grep (job_id|sys\.jobs|async_job|wait.*index|poll) across deno code returns only that one README line — no polling exists in schema.ts). setupSchema issues CREATE INDEX ASYNC and returns immediately.


Either implement a job-status poll loop, or soften the README line to something like: "setupSchema issues CREATE INDEX ASYNC — DSQL builds the index asynchronously in the background; reads work immediately and benefit from the index as DSQL promotes it." I'd pick the soften: polling in a sample's setup path adds plumbing without teaching value.

.github/workflows/deno-booking-api-integ-tests.yml
```diff
@@ -0,0 +1,64 @@
...
+      - 'deno/booking-api/postgres-js/**'
+      - '.github/workflows/deno-booking-api-integ-tests.yml'
+
+jobs:
+  create-cluster:
+    uses: ./.github/workflows/dsql-cluster-create.yml
+    with:
+      workflow_name: deno-booking-api
+    secrets:
+      AWS_IAM_ROLE: ${{ secrets.JAVASCRIPT_IAM_ROLE }}
```
amaksimo, line 20: yea do not add DENO_IAM_ROLE, pointless, this is easier to follow anyway (from your questions)

.github/workflows/deno-booking-api-integ-tests.yml
```diff
@@ -0,0 +1,64 @@
+name: Deno Booking API integration tests
```
amaksimo, line 1: I think we can combine the CIs here. Since we are just testing this sample, no need for two IMO.

deno/booking-api/postgres-js/schema.ts
```diff
@@ -0,0 +1,241 @@
...
+    // DDL 4: non-admin role (no IF NOT EXISTS support → catch 42710).
+    //
+    // NOTE: This sample creates `non_admin_user` with CRUD privileges but
+    // the runtime pool still connects as `CLUSTER_USER` (typically admin).
+    // In production, map `non_admin_user` to an IAM role and run CRUD
+    // through that client; reserve the admin client for DDL/setup only.
+    // See the Aurora DSQL docs on database roles and IAM authentication:
+    // https://docs.aws.amazon.com/aurora-dsql/latest/userguide/using-database-and-iam-roles.html
+    await ignoreSqlState("42710", () =>
+      sql.unsafe(`CREATE ROLE non_admin_user`)
```
amaksimo, line 128: we create this but from what I can tell do not use it?

deno/booking-api/postgres-js/handlers.ts
```diff
@@ -0,0 +1,617 @@
...
+    }
+  }
+
+  const timeRange = parseTimeRange(body.start_time, body.end_time);
+  if ("error" in timeRange) return timeRange.error;
+
+  try {
+    const result = await withOccRetry(() =>
+      ctx.sql.begin(async (tx) => {
+        // Overlap detection inside the transaction. The unique index on
```
amaksimo: The existing comment says the unique index is the "commit-time backstop," but it only catches identical windows (same resource + same start + same end). Two concurrent transactions inserting overlapping but distinct windows (e.g., 9:00–10:00 and 9:30–10:30) will both pass this SELECT (both see 0 rows) and both commit successfully — DSQL's row-based OCC won't conflict them because they wrote different rows.

The README's "Concurrency model" section documents this, but since this is the code people will copy for real booking systems, I'd suggest making the limitation more visible right here. Something like:

// ⚠️ CONCURRENT-WRITE LIMITATION: This overlap check prevents
// double-booking for sequential writes. Under concurrent writes,
// two transactions with overlapping-but-distinct windows can both
// pass this SELECT (both see 0 rows) and both commit — DSQL's
// row-based OCC only conflicts transactions that touch the same
// physical rows. The unique index catches identical windows only.
//
// For strict serialization, add a `resources` table and acquire
// SELECT ... FOR UPDATE on the resource row before the overlap
// check. See README § "Concurrency model" for details.
This way anyone reading the handler code gets the caveat immediately, without needing to find the right README section.

deno/booking-api/postgres-js/schema.ts
gxjx-x: yes we do not use it, mentioned in line 122, i think we want to share the best practice here, i don't see a better way of documenting this. I do not want to complicate the sample by having to set up the IAM roles manually.

.github/workflows/deno-booking-api-integ-tests.yml
gxjx-x: I like it this way because, it keeps the CI and sample 1:1 allowing both to evolve independently, just in case if they ever do. also i think its cleaner for now, we can consider merging this in the future.

deno/booking-api/postgres-js/handlers.ts
gxjx-x: resolved in https://github.com/aws-samples/aurora-dsql-samples/pull/1031/changes/72443351e5cb7755be65c92953b458b9426a9746

deno/booking-api/postgres-js/README.md
gxjx-x: https://github.com/aws-samples/aurora-dsql-samples/pull/1031/changes/3ebc8ebb26025d17c57e96e5d032adae2ada6cd6

deno/booking-api/postgres-js/README.md
gxjx-x: resolved: https://github.com/aws-samples/aurora-dsql-samples/pull/1031/changes/3ebc8ebb26025d17c57e96e5d032adae2ada6cd6

deno/booking-api/postgres-js/README.md
gxjx-x: https://github.com/aws-samples/aurora-dsql-samples/pull/1031/changes/3ebc8ebb26025d17c57e96e5d032adae2ada6cd6

deno/booking-api/postgres-js/README.md
gxjx-x: https://github.com/aws-samples/aurora-dsql-samples/pull/1031/changes/c1f270905e7df674924e47e202ba68b4148b8e03

deno/booking-api/postgres-js/README.md
gxjx-x: https://github.com/aws-samples/aurora-dsql-samples/pull/1031/changes/c1f270905e7df674924e47e202ba68b4148b8e03

deno/booking-api/postgres-js/occ-retry.ts
gxjx-x: https://github.com/aws-samples/aurora-dsql-samples/pull/1031/changes/47746a264d70fe6a433d600b55b455004d00936b

deno/booking-api/postgres-js/occ-retry.ts
gxjx-x: https://github.com/aws-samples/aurora-dsql-samples/pull/1031/changes/3ebc8ebb26025d17c57e96e5d032adae2ada6cd6

deno/booking-api/postgres-js/README.md
gxjx-x: good catch i missed this. resolved in: https://github.com/aws-samples/aurora-dsql-samples/pull/1031/changes/3ebc8ebb26025d17c57e96e5d032adae2ada6cd6 with other clean up.

deno/booking-api/postgres-js/README.md
gxjx-x: removed, fixed in: https://github.com/aws-samples/aurora-dsql-samples/pull/1031/changes/47746a264d70fe6a433d600b55b455004d00936b

deno/booking-api/postgres-js/README.md
gxjx-x: trimmed this in : https://github.com/aws-samples/aurora-dsql-samples/pull/1031/changes/3ebc8ebb26025d17c57e96e5d032adae2ada6cd6

PR #1216, opened 2026-06-01
feat: add Amazon Aurora DSQL auth and session management sample

charycha:
Adds a Node.js + TypeScript sample application demonstrating user authentication and session management on Amazon Aurora DSQL.

The sample shows how to:
- Use IAM-based database authentication via the official DSQL connector
- Generate UUIDs for primary keys in the application layer
- Enforce application-level referential integrity (no foreign keys)
- Implement OCC retry logic with exponential backoff and jitter for serialization errors (SQLSTATE 40001)
- Use CREATE INDEX ASYNC for non-blocking index creation
- Run each DDL statement in its own transaction
- Hash session tokens with SHA-256 (plaintext returned to client once)
- Hash passwords with bcrypt and unique salts in the application layer

Stack: Express.js, node-postgres via @aws/aurora-dsql-node-postgres-connector, Vitest for tests.

### DIFF
--- a/sample-amazon-aurora-dsql-auth-session-mgmt/.gitignore
+++ b/sample-amazon-aurora-dsql-auth-session-mgmt/.gitignore
@@ -1,0 +1,36 @@
+# Dependencies
+node_modules/
+
+# Build output
+dist/
+build/
+
+# Environment files
+.env
+.env.local
+.env.*.local
+
+# Logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# Blog assets
+blog/
+
+# Placeholder files
+.gitkeep
+
+# Editor / OS
+.DS_Store
+.vscode/
+.idea/
+*.swp
+
+# Test coverage
+coverage/
+.nyc_output/
+
+# TypeScript
+*.tsbuildinfo
--- a/sample-amazon-aurora-dsql-auth-session-mgmt/README.md
+++ b/sample-amazon-aurora-dsql-auth-session-mgmt/README.md
@@ -1,12 +1,12 @@
 # User Authentication Service with Session Management on Aurora DSQL
 
-A complete user authentication service with session management built on Amazon Aurora DSQL. Demonstrates DSQL-specific patterns including strong consistency, IAM authentication, OCC retry, and application-level referential integrity.
+A user authentication service with session management built on Amazon Aurora DSQL. Demonstrates DSQL-specific patterns including strong consistency, IAM authentication, OCC retry, and application-level referential integrity.
 
 ## Tech Stack
 
 - Node.js 20+ / TypeScript
 - Express.js
-- Amazon Aurora DSQL (via @aws/aurora-dsql-node-postgres-connector)
+- Amazon Aurora DSQL (via `@aws/aurora-dsql-node-postgres-connector`)
 - Vitest for testing
 
 ## Prerequisites
@@ -54,12 +54,12 @@ npm test
 # 1. Register
 curl -s -X POST http://localhost:3000/api/auth/register \
   -H "Content-Type: application/json" \
-  -d '{"email": "alice@test.com", "password": "secureP@ss1"}' | jq
+  -d '{"email": "alice@example.com", "password": "secureP@ss1"}' | jq
 
 # 2. Login
 curl -s -X POST http://localhost:3000/api/auth/login \
   -H "Content-Type: application/json" \
-  -d '{"email": "alice@test.com", "password": "secureP@ss1"}' | jq
+  -d '{"email": "alice@example.com", "password": "secureP@ss1"}' | jq
 
 # 3. Get profile (use token from login)
 curl -s http://localhost:3000/api/auth/me \
@@ -85,17 +85,17 @@ curl -s http://localhost:3000/api/auth/me \
 # Duplicate registration (expect 409)
 curl -s -X POST http://localhost:3000/api/auth/register \
   -H "Content-Type: application/json" \
-  -d '{"email": "alice@test.com", "password": "secureP@ss1"}' | jq
+  -d '{"email": "alice@example.com", "password": "secureP@ss1"}' | jq
 
 # Wrong password (expect 401, generic message)
 curl -s -X POST http://localhost:3000/api/auth/login \
   -H "Content-Type: application/json" \
-  -d '{"email": "alice@test.com", "password": "wrongpassword"}' | jq
+  -d '{"email": "alice@example.com", "password": "wrongpassword"}' | jq
 
-# Non-existent email (expect same 401 message - user enumeration prevention)
+# Non-existent email (expect same 401 message — user enumeration prevention)
 curl -s -X POST http://localhost:3000/api/auth/login \
   -H "Content-Type: application/json" \
-  -d '{"email": "nobody@test.com", "password": "secureP@ss1"}' | jq
+  -d '{"email": "nobody@example.com", "password": "secureP@ss1"}' | jq
 
 # Invalid email format (expect 400)
 curl -s -X POST http://localhost:3000/api/auth/register \
@@ -105,7 +105,7 @@ curl -s -X POST http://localhost:3000/api/auth/register \
 # Short password (expect 400)
 curl -s -X POST http://localhost:3000/api/auth/register \
   -H "Content-Type: application/json" \
-  -d '{"email": "new@test.com", "password": "short"}' | jq
+  -d '{"email": "new@example.com", "password": "short"}' | jq
 
 # Missing fields (expect 400)
 curl -s -X POST http://localhost:3000/api/auth/register \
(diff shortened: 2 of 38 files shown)


amaksimo:
# PR Self-Review — Validated Findings

Reviewed at head SHA `550a240f99d37aab162aa7ef44c3188fd00cdf9a`. Each finding below was verified against the actual file contents at this SHA. Items marked **Disposition** as `false-positive` or `softened` are notes from a sequential re-verification pass.

| # | Confidence | Area | Finding | Disposition |
|---|---|---|---|---|
| 1 | 98 | Security / IDOR | `revokeSession(userId, sessionId)` accepts `userId` but discards it; calls `revokeById(sessionId)` which only filters `WHERE id = $1`. Any authenticated user can revoke any other user's session. | open — release-blocking |
| 2 | 90 | Security / Enumeration | Login no-user branch returns immediately without performing a bcrypt verify; the wrong-password branch performs one bcrypt verify (~80–100 ms at cost 10). Identical messages but ~80–100 ms timing delta defeats the documented anti-enumeration property. | open |
| 3 | 90 | DSQL / Schema | `sessions.token_hash VARCHAR(64) NOT NULL UNIQUE` already produces a backing unique index in DSQL (per DSQL `ALTER TABLE … ADD table_constraint_using_index` and `CREATE INDEX ASYNC` docs). The explicit `CREATE INDEX ASYNC IF NOT EXISTS idx_sessions_token_hash` then creates a second redundant index on the same column. | open |
| 4 | 60 | Comment rot | `app.ts:11-12` says the default-exported bare app is "kept for backward compatibility with existing tests" — but this is a brand-new sample. Re-verification: the export *is* consumed by `app.test.ts:4`, so it is not dead code. **Softened** to comment-rot only — drop the "backward compatibility" phrasing. | softened |
| 5 | 85 | Tests / Tautological | `sessionService.test.ts` `revokeSession` test only asserts `sessionRepository.revokeById` was called with the forwarded `sessionId`. There is no behavioral test that `revokeSession` rejects revocation of another user's session — which is exactly why finding #1 was not caught. | open |
| 6 | 85 | Error handling | `retryWithBackoff.ts:122-155` captures `lastError` but the final `throw new ServiceUnavailableError(…)` at L153 has no `cause` and no log of `lastError`. Operators receive a generic 503 with zero forensic context about which SQLSTATE / query drove the OCC storm. | open |
| 7 | 80 | Type safety | `app.ts:37-39` declares `authMiddleware: any` with `eslint-disable @typescript-eslint/no-explicit-any`, even though `RequestHandler` is already imported on L19. Defeats type-checking on a security-critical dependency. | open |
| 8 | 80 | Doc / DSQL | `migrate.ts` finishes before its ASYNC indexes are VALID. DSQL `CREATE INDEX ASYNC` returns immediately and builds in background. First reads after a fresh migration will full-scan `sessions` until `idx_sessions_user_id` and `idx_sessions_token_hash` become valid. README and the `migrate.ts` header don't mention this. | open |
| 9 | 80 | Comment rot | `sessionRepository.ts:315` comment says "parses the JSONB `client_metadata` column" but `migrate.ts:64` declares it `TEXT`. Behavior is correct (`parseClientMetadata` handles both shapes); comment is wrong. | open |
| 10 | 80 | UX / DX | `auth.ts:93-101` uses identical message `'Authorization header is required'` for both the "no header / wrong scheme" branch and the "header present but token empty after `Bearer `" branch. The second message is misleading — the header *is* present. | open |
| 11 | 75 | Error handling | `userRepository.create` (L112-120), `sessionRepository.{create, revokeById, revokeAllByUserId}`, and `migrate.ts:97-104` all use `} catch (error) { await client.query('ROLLBACK'); … throw error; }`. If `ROLLBACK` itself throws (e.g. dead connection — the typical reason the original query threw), control exits with the ROLLBACK error and the original SQLSTATE is lost. In `userRepository.create`, this means `isUniqueViolation(error)` (L116) never runs and a duplicate-email registration could surface as an opaque 500 instead of `ConflictError`. | open |
| 12 | 75 | Resilience | `sessionRepository.ts:356-361` `getRowCount` returns `0` when `result.rowCount` is absent. The `do { … } while (batchRevoked === 3000)` loop at L300 is load-bearing on this value — a missing `rowCount` exits the loop after one iteration and `revokeAllByUserId` returns `0` even when rows were updated. Today's `pg` driver always sets `rowCount`; the silent fallback is fragile to driver swap or test mocks. | open |
| 13 | 75 | Security / Hardening | `grep` confirms no `express-rate-limit`, no `rateLimit`, and no `trust proxy` configuration anywhere in the sample. `routes/authRoutes.ts:119` records `req.ip` into session metadata — behind any AWS load balancer this will be the LB address; if a future operator naively sets `app.set('trust proxy', true)`, `X-Forwarded-For` becomes spoofable. README markets this as a security pattern. | open |
| 14 | 60 | Security / Defaults | `passwordHasher.ts:31` sets `COST_FACTOR = 10`. **Softened** because L29 explicitly says "good balance between security and latency for a proof-of-concept" — the value is a documented choice. Worth a one-line README note to bump for production use, but not a bug. | softened |
| 15 | 70 | Silent failure | `sessionRepository.ts:335-347` `parseClientMetadata` swallows `JSON.parse` errors with an empty `catch { return {} }`. No log, no error id. Any future direct-SQL writer or row-corruption scenario produces silently empty metadata that is indistinguishable from "no metadata". | open |
| 16 | 100 | Dependencies | `package.json:36` declares `fast-check@^4.7.0` in devDependencies. `grep -rn "fast-check\|fc\." src/` returns zero results. Either remove the dependency or add at least one property test. | open |

## Items checked and **not** flagged (kept here so reviewers don't re-check)

- DSQL DDL syntax: `dsql_lint` clean on all four statements at `migrate.ts`.
- DSQL DML: `INSERT … RETURNING`, `UPDATE … WHERE id IN (SELECT … LIMIT $N)` — both lint clean and supported.
- SAVEPOINT — not used; consistent with DSQL.
- OCC SQLSTATE 40001 handling — matches DSQL guidance.
- `crypto.randomBytes(32)` token entropy — adequate.
- SHA-256 of high-entropy session tokens — adequate.
- All SQL is parameterized — no injection surface.
- README's strong-consistency claim about session revocation — accurate per DSQL docs.
- Email validator "reads original value before normalization" — `validateEmail` re-normalizes internally, so this is **not** a bug.
- Retry log message off-by-one (`attempt + 1 / maxRetries + 1`) — correct.
- `errorHandler` 4-arg signature — correct for Express.

## Suggested triage

- **Block on:** #1 (IDOR), #2 (timing), #3 (duplicate index), #5 (tautological tests that masked #1).
- **Fix-before-merge:** #6, #7, #8, #9, #10, #16.
- **Follow-up nice-to-haves:** #4, #11, #12, #13, #14, #15.

No code or tests were modified by this review.

charycha:
@amaksimo thanks for the thorough review- every finding was actionable. Pushed commit 33c6392 addressing all 16 in a single batch (per-finding mapping in the commit message). Quick highlights:

Security blockers:
- #1 (IDOR): repository method renamed to `revokeByIdForUser` with `WHERE id = $1 AND user_id = $2`; service throws `InvalidSessionError` when zero rows match. Behavioral test in `sessionService.test.ts` covers the cross-user reject case explicitly.
- #2 (timing): added `dummyVerify()`; no-user branch now runs bcrypt against a fixed dummy hash. Regression test asserts `bcrypt.compare` is called even when the email is unknown.
- #3 (duplicate index): removed `idx_sessions_token_hash`; the `UNIQUE` constraint already creates the backing index.
- #5: tautological test replaced with the behavioral test in #1.

Fix-before-merge:all addressed (#6 cause + log, #7 typed, #9 comment, #10 three distinct messages, #11 best-effort rollback in 5 places, #12 strict rowCount, #16 fast-check removed).

Follow-ups: README now has a Production Hardening Checklist covering rate limiting, trust proxy, bcrypt cost, structured logging, and cookie storage (#8, #13, #14). #4 and #15 fixed inline. #8 also has an inline note in `migrate.ts`.

amaksimo:
## Re-review against `33c6392` — all 16 findings resolved ✅

Verified each finding from the [previous review](https://github.com/aws-samples/aurora-dsql-samples/pull/1216#issuecomment-4607273564) against the files at commit `33c6392`. All resolved, no regressions spotted.

| # | Finding | Status | Verification |
|---|---|---|---|
| 1 | IDOR on `revokeSession` | ✅ Fixed | `sessionRepository.ts` `revokeByIdForUser` filters `WHERE id=$1 AND user_id=$2`; service throws `InvalidSessionError` on zero rows. |
| 2 | Login timing side-channel | ✅ Fixed | `passwordHasher.ts:dummyVerify` + `DUMMY_HASH`; `authService.ts:155` calls it in the no-user branch. |
| 3 | Duplicate index on `sessions.token_hash` | ✅ Fixed | `migrate.ts` only creates `idx_sessions_user_id`; comment now explains the UNIQUE backing index. |
| 4 | Comment rot on default-exported `app` | ✅ Fixed | `app.ts` rewords as "bare app for app-level tests"; matches actual usage in `app.test.ts`. |
| 5 | Tautological session tests | ✅ Fixed | `sessionService.test.ts` adds real cross-user IDOR rejection tests. |
| 6 | `lastError` discarded by `retryWithBackoff` | ✅ Fixed | `retryWithBackoff.ts:157-171` logs structured `console.error` and `throw new ServiceUnavailableError(..., { cause: lastError })`. |
| 7 | `authMiddleware: any` in `AppDependencies` | ✅ Fixed | `app.ts:44` now `authMiddleware: RequestHandler`. |
| 8 | Missing CREATE INDEX ASYNC documentation | ✅ Fixed | `migrate.ts:69-77` adds the async-validity caveat with docs link. |
| 9 | JSONB → TEXT comment rot | ✅ Fixed | `sessionRepository.ts:354` comment now clarifies TEXT (no JSONB). |
| 10 | Identical "Bearer required" error message | ✅ Fixed | `auth.ts:93-111` distinguishes missing header / wrong scheme / empty token. |
| 11 | ROLLBACK clobbers original error | ✅ Fixed | Inner `try { ROLLBACK } catch {}` then `throw error` in `userRepository.ts`, `sessionRepository.ts` (×3), `migrate.ts`. |
| 12 | Silent fallback in `getRowCount` | ✅ Fixed | `sessionRepository.ts:405-412` now throws when `rowCount` missing. |
| 13 | No rate limiting | ✅ Documented | README "Production hardening checklist" calls out `express-rate-limit`. |
| 14 | bcrypt cost factor 10 | ✅ Documented | README + `passwordHasher.ts` flag this as PoC; recommend ≥12 in production. |
| 15 | `parseClientMetadata` swallows JSON errors | ✅ Fixed | `sessionRepository.ts:383-389` logs `console.warn` with the parse error before defaulting to `{}`. |
| 16 | `fast-check` declared but unused | ✅ Fixed | Removed from `package.json` devDependencies (no remaining references). |

### Bonus improvements introduced by `33c6392`
- `AppError` / `ServiceUnavailableError` now accept the standard `ErrorOptions`, enabling proper `Error.cause` chains.
- ROLLBACK preservation pattern is now consistent across every transactional code path (good defensive baseline).
- README "Production hardening checklist" calls out trust-proxy / `X-Forwarded-For` spoofing — important since `req.ip` is persisted in `client_metadata`.

No remaining blocking issues. LGTM to merge once CI is green.

amaksimo (approved):

charycha:
Pushed commit 8af7737 — repo is now in lockstep with the blog updates.
Mapping back to the comments in the latest review:

#5 (maxLifetimeSeconds: 3300): added to connection.ts pool config.

#9 (use connector's OCC retry): repositories now route transactional
work through pool.transaction() when available (production: AuroraDSQLPool
from @aws/aurora-dsql-node-postgres-connector), falling back to a manual
BEGIN/COMMIT wrapped in retryWithBackoff for unit-test mocks. Eliminates
the duplicated retry plumbing in every repository method.

#4 (sys.wait_for_job for async indexes): migrate.ts now captures the
job_id from CREATE INDEX ASYNC and SELECT sys.wait_for_job($1)s before
returning. Opt-out flag for callers that want non-blocking migrations.

#7 (housekeeping for expired/revoked rows): new src/scripts/housekeeping.ts
with a batched, idempotent DELETE loop respecting the 3000-row cap.
Wired up as `npm run housekeeping`. Configurable retention via
SESSION_RETENTION_DAYS env, defaulting to 30 days.

End-to-end validated against a live DSQL cluster before pushing — the
register/login/profile/list/revoke flow works, the IDOR fix from the
previous round still rejects cross-user revocation, and the schema
empirics (sessions_token_hash_key auto-created from UNIQUE) match the
post.

Tests: 197/197 passing.

PR #1258, opened 2026-06-08
docs(typescript/sequelize): use native JSONB instead of TEXT workaround

amaksimo:
## Summary

DSQL now supports `JSON` and `JSONB` column types natively, so the Sequelize sample's TEXT-with-getter/setter serialization workaround is obsolete. Replace it with a plain `DataTypes.JSONB` column.

Verified against a live cluster: `JSONB` column creation, `'{...}'::jsonb` casts, `->`/`->>` operators, and `jsonb_array_length` all work end-to-end.

## Test plan

- [x] Confirmed JSONB roundtrip on a live DSQL cluster.
- [ ] Build the sequelize sample to make sure the README's snippet stays consistent with the codebase (no code changes were needed).


praba2210 (approved):
looks good.

PR #1363, opened 2026-06-27
feat(python): Add Tortoise ORM rideshare sample for Aurora DSQL

lasitab-code:
Body:
  
  ## Summary
  Adds a Python sample demonstrating Tortoise ORM with Aurora DSQL, including:
  - Async CRUD operations for a rideshare domain (riders, drivers, rides, payments)
  - IAM authentication (admin + non-admin user flows)
  - OCC retry with exponential backoff and jitter
  - asyncpg connection pool patching for DSQL compatibility
  - Schema isolation for multi-tenant patterns
  - Unit tests (pytest)
  
  ## Testing
  - End-to-end tested with admin and non-admin users against a live DSQL cluster (us-west-2)
  - Unit tests pass: `python -m pytest` (5 passed)
  - Security scan: 0 CVEs, 0 secrets, 0 code vulnerabilities (Trivy + Gitleaks)


python/tortoise-orm/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+If your production architecture uses IAM roles (EC2 instance profiles, ECS task roles, Lambda execution roles), create a dedicated IAM role with `dsql:DbConnect` permission scoped to your cluster. The same database linking pattern applies — use `CREATE ROLE` with a simple name, then `AWS IAM GRANT` to map it to the IAM role ARN, and `GRANT` the schema permissions. See the "Advanced" section in `src/setup_app_user.sql` for an example.
+
+## Project structure
+
+```
+.
+├── pyproject.toml          # Project metadata and pytest configuration
+├── requirements.txt        # Python dependencies
+├── .env.example            # Environment variable template
```
praba2210: Lists `.env.example`, but no such file is in the PR and the Configuration section uses `export` statements, not a dotenv file. Remove this line or add the file.

python/tortoise-orm/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+- No static passwords — IAM tokens are generated at runtime via boto3
+- TLS required for all connections to Aurora DSQL
+- Tokens expire after 15 minutes; regenerate for long-running applications
+
+## Related
+
+- [Aurora DSQL User Guide](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/what-is-aurora-dsql.html)
+- [Tortoise ORM Documentation](https://tortoise.github.io/)
+- [Blog post: Building async Python applications with Tortoise ORM and Aurora DSQL](https://aws.amazon.com/blogs/database/)
```
praba2210: Links to the generic AWS Database blog index rather than a real post — placeholder URL. Remove the bullet or point it at the actual article.

python/tortoise-orm/src/rider_config.py
```diff
@@ -0,0 +1,174 @@
...
+
+if not CLUSTER_USER:
+    raise ValueError(
+        "Missing required environment variable CLUSTER_USER. "
+        "Set to 'admin' for admin access or your app role name (e.g., 'rideshare_app') for non-admin."
+    )
+
+# Admin uses "public" schema; non-admin users use a custom schema
+ADMIN_USER = "admin"
+SCHEMA = "public" if CLUSTER_USER == ADMIN_USER else os.environ.get("CLUSTER_SCHEMA", "myschema")
```
praba2210: Non-admin `SCHEMA` defaults to `"myschema"`, but the README and `setup_app_user.sql` create/use `"rideshare"`. If a reader skips the `CLUSTER_SCHEMA` export, the app targets a schema that was never created or granted. Consider defaulting to `"rideshare"` or requiring the env var for non-admin. (The `myschema` default matches the sqlalchemy sibling, so this is an internal doc/default mismatch, not a wrong default.)

python/tortoise-orm/src/rider_config.py
```diff
@@ -0,0 +1,174 @@
...
+            CLUSTER_ENDPOINT, CLUSTER_REGION
+        )
+    else:
+        token = client.generate_db_connect_auth_token(
+            CLUSTER_ENDPOINT, CLUSTER_REGION
+        )
+    return token
+
+
+async def _dsql_safe_reset(self, *, timeout=None):
```
praba2210, line 48: This replaces asyncpg's `reset()` wholesale, which also drops its conditional `ROLLBACK`. No effect here since every op runs in autocommit, but a copier who wraps work in an explicit transaction could get a pooled connection reused mid-transaction. Worth a comment noting the autocommit assumption.

python/tortoise-orm/src/rider_config.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+async def create_schema():
+    """Create database tables one at a time using a raw asyncpg connection.
+
+    Aurora DSQL does not support multiple DDL statements in a single
+    transaction. Each CREATE TABLE executes as its own statement.
+    We use a raw asyncpg connection (not the Tortoise pool) to avoid
+    pool reset issues during DDL execution."""
+    ssl_ctx = ssl_module.create_default_context()
```
praba2210: The SSL context, IAM token, and connection params are rebuilt here, duplicating `init_db()` (line 75+). Extracting a shared helper avoids the two copies drifting.

python/tortoise-orm/src/retry.py
```diff
@@ -0,0 +1,67 @@
...
+T = TypeVar("T")
+BASE_DELAY = 0.05  # 50ms
+
+# SQLSTATE code for serialization failure (OCC conflict)
+OCC_SQLSTATE = "40001"
+
+
+def _is_occ_error(e: Exception) -> bool:
+    """Check if an exception is an OCC conflict using structured SQLSTATE inspection."""
+    if isinstance(e, PostgresError) and hasattr(e, 'sqlstate'):
```
praba2210: `hasattr(e, 'sqlstate')` is redundant — every `asyncpg.PostgresError` exposes `sqlstate`. Can simplify to `if isinstance(e, PostgresError):`.

python/tortoise-orm/src/retry.py
```diff
@@ -0,0 +1,67 @@
...
+        return e.sqlstate == OCC_SQLSTATE
+    # Fallback for wrapped exceptions (e.g., Tortoise OperationalError)
+    error_msg = str(e)
+    return OCC_SQLSTATE in error_msg or "OC000" in error_msg or "OC001" in error_msg
```
praba2210: This substring fallback can false-positive on unrelated errors whose message happens to contain `40001`/`OC000`/`OC001`, causing a wrongful retry. Marginal since the primary sqlstate check covers the common case, but worth narrowing.

python/tortoise-orm/src/riders_app.py
```diff
@@ -0,0 +1,246 @@
...
+
+    # OCC retry demonstration
+    print("\n▶ OCC retry pattern...")
+    await demonstrate_occ_retry()
+
+    # Summary
+    print("\n" + "=" * 70)
+    total_rides = await Ride.filter(status="completed").count()
+    total_revenue = sum(
+        [r.fare_amount for r in await Ride.filter(status="completed")]
```
praba2210: List comprehension inside `sum()` builds a throwaway list — drop the brackets for a generator: `sum(r.fare_amount for r in await Ride.filter(status="completed"))`.

python/tortoise-orm/requirements.txt
```diff
@@ -0,0 +1,6 @@
+# Application dependencies
+tortoise-orm[asyncpg]==1.1.7
+boto3>=1.43.36
+
+# Testing
+pytest>=8.0.0
```
praba2210, line 6: No `pytest-asyncio`, so the async core (`with_retry`, `create_schema`, `main`) can't be exercised by tests — the current suite only covers config parsing and existence checks. Consider adding `pytest-asyncio` plus an async smoke test.

praba2210 (commented):
## Tortoise ORM rideshare sample

Reviewed at head `72c51cf`. No correctness blocker — the sample runs and its security posture is sound (verified: DSQL supports inline `UNIQUE`, Tortoise maps the `schema` credential to `search_path`, `SET search_path` is regex-guarded, TLS is verify-full, IAM tokens are never logged). 

#### Left inline comment and below 2 are CI & Readme

1. **No CI workflow.** There's no `python-tortoise-integ-tests.yml` and no `ci-gate.yml` path filter, so the sample gets no automated coverage. Most samples have one; `python/sqlalchemy`, `python/jupyter`, `java/liquibase`, `typescript/drizzle`, and `lambda/sample` also lack it, so this isn't an enforced gate — but adding one would match the majority convention.
2. **Not in the root README samples table.** The sample is missing from the top-level `README.md` index (as are `sqlalchemy`, `drizzle`, and `prisma-multi-region`). A one-line row would aid discoverability.

python/tortoise-orm/README.md
lasitab-code: Fixed. Removed .env.example from the project structure listing.

python/tortoise-orm/src/rider_config.py
lasitab-code: Fixed. Changed the default from "myschema" to "rideshare" to match the README and setup_app_user.sql.

python/tortoise-orm/README.md
lasitab-code: Fixed. Removed the placeholder blog link.

python/tortoise-orm/src/rider_config.py
lasitab-code: Added a comment documenting the autocommit assumption and noting the risk if explicit transactions are used without Tortoise's transaction context manager.

python/tortoise-orm/src/rider_config.py
lasitab-code: Fixed --  extracted _connection_params() helper that both init_db() and create_schema() now .

python/tortoise-orm/src/retry.py
lasitab-code: Removed — simplified to if isinstance(e, PostgresError) .

python/tortoise-orm/src/retry.py
lasitab-code: Narrowed to "SQLSTATE 40001" in error_msg plus DSQL-specific OC000/OC001 codes.

python/tortoise-orm/requirements.txt
lasitab-code: Added pytest-asyncio>=0.23.0 and async smoke tests for with_retry.

python/tortoise-orm/src/riders_app.py
lasitab-code: Fixed — using generator expression now , dropped the bracket

praba2210:
LGTM!

praba2210 (approved):

PR #1492, opened 2026-07-17
fix: Aurora DSQL auth-session-mgmt sample against current DSQL behaviour

charycha:
## Summary

Two fixes to `sample-amazon-aurora-dsql-auth-session-mgmt/` surfaced by running the sample end-to-end against a fresh Aurora DSQL cluster.

### 0314671 - fix: use CALL for sys.wait_for_job

Aurora DSQL defines `sys.wait_for_job` as a procedure, not a function. The migrator's `SELECT sys.wait_for_job($1)` call fails at startup with:

This blocks `npm start` on any fresh DSQL cluster because migrations run at boot. Fixed by changing the invocation to `CALL`.

### 5d4c907 - feat: SKIP_MIGRATIONS env var for non-admin runtime

The README's Production hardening checklist recommends running the app as a least-privilege `app_runtime` role and keeping `admin` for setup. That guidance is unimplementable today because the app calls `runMigrations(pool)` unconditionally on every startup. When the app runs as `app_runtime`, that call fails with `permission denied for schema public` (SQLSTATE 42501). Aurora DSQL doesn't support schema-level `GRANT` (`feature not supported on system entity`), so no privilege-only workaround exists.

Added a `SKIP_MIGRATIONS=true` env var that skips `runMigrations` when set. Also documented in the README's Production hardening section.

## Testing

Ran the full sample end-to-end against a fresh single-Region DSQL cluster in us-east-1: cluster create → migrations (as admin) → all six API endpoints exercised → OCC retry demonstrated via concurrent revoke-all → runtime role setup and switch (with `SKIP_MIGRATIONS=true`) → housekeeping purge → cleanup.

### DIFF
--- a/sample-amazon-aurora-dsql-auth-session-mgmt/src/db/migrate.ts
+++ b/sample-amazon-aurora-dsql-auth-session-mgmt/src/db/migrate.ts
@@ -185,8 +185,8 @@ export async function runMigrations(
     if (waitForAsyncJobs && asyncJobId) {
       const waitClient = await pool.connect();
       try {
-        // DSQL changed sys.wait_for_job from a function to a procedure,
-        // so it must be invoked with CALL, not SELECT.
+        // sys.wait_for_job is a procedure, not a function in DSQL, so it
+        // must be invoked with CALL, not SELECT.
         await waitClient.query('CALL sys.wait_for_job($1)', [asyncJobId]);
       } finally {
         waitClient.release();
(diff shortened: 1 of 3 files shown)


sample-amazon-aurora-dsql-auth-session-mgmt/src/db/migrate.ts
```diff
@@ -185,7 +185,9 @@ export async function runMigrations(
     if (waitForAsyncJobs && asyncJobId) {
       const waitClient = await pool.connect();
       try {
-        await waitClient.query('SELECT sys.wait_for_job($1)', [asyncJobId]);
+        // DSQL changed sys.wait_for_job from a function to a procedure,
+        // so it must be invoked with CALL, not SELECT.
```
amaksimo: I think this comment should be changed to something like `sys.wait_for_job is a prodecure, not function in DSQL so we must invoke with CALL, not SELECT`

amaksimo (approved):

PR #1704, opened 2026-08-24
Fix Rails primary key introspection for DSQL

amaksimo:
## Summary

* temporarily override Rails 7.2 primary-key introspection to read `pg_constraint.conkey`
* add bulk-upsert regression coverage to both Rails samples
* correct Petclinic schema primary-key declarations generated by the old query
* pin Minitest below 6 for Rails 7.2 test compatibility
* document the workaround in the Rails sample README

## Testing

* Petclinic: 1 run, 1 assertion, 0 failures
* Car Rental: 13 runs, 33 assertions, 0 failures
* DSQL SQL lint: 0 errors, 0 warnings

### DIFF
--- a/ruby/rails/README.md
+++ b/ruby/rails/README.md
@@ -73,6 +73,23 @@ require "active_record/connection_adapters/postgresql/schema_statements"
 module ActiveRecord::ConnectionAdapters::PostgreSQL::SchemaStatements
   # Override client_min_messages for DSQL compatibility
   def client_min_messages=(level); end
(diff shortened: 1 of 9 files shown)


amaksimo:
Superseded by #1705, which contains the same commit on an upstream feature branch so the repository-secret-backed DSQL integration tests can run. The failures here occurred before tests because fork PRs cannot access `RUBY_IAM_ROLE`; they were not code failures.

PR #1705, opened 2026-08-24
Fix Rails primary key introspection for DSQL

amaksimo:
## Summary

* temporarily override Rails 7.2 primary-key introspection to read `pg_constraint.conkey`
* add bulk-upsert regression coverage to both Rails samples
* correct Petclinic schema primary-key declarations generated by the old query
* pin Minitest below 6 for Rails 7.2 test compatibility
* document the workaround in the Rails sample README

## Testing

* Petclinic: 1 run, 1 assertion, 0 failures
* Car Rental: 13 runs, 33 assertions, 0 failures
* DSQL SQL lint: 0 errors, 0 warnings

### DIFF
--- a/.github/workflows/ruby-rails-integ-tests.yml
+++ b/.github/workflows/ruby-rails-integ-tests.yml
@@ -49,7 +49,7 @@ jobs:
         CLUSTER_ENDPOINT: ${{ needs.create-cluster.outputs.cluster-endpoint }}
         REGION: ${{ needs.create-cluster.outputs.region }}
         CLUSTER_USER: admin
-        RAILS_ENV: development
+        RAILS_ENV: test
       run: |
         bundle install
         mkdir -p ~/.postgresql
--- a/ruby/rails/README.md
+++ b/ruby/rails/README.md
@@ -73,6 +73,23 @@ require "active_record/connection_adapters/postgresql/schema_statements"
 module ActiveRecord::ConnectionAdapters::PostgreSQL::SchemaStatements
   # Override client_min_messages for DSQL compatibility
   def client_min_messages=(level); end
+
+  # Temporary workaround until the Rails pg_constraint fix is released
+  def primary_keys(table_name)
+    query_values(<<~SQL, "SCHEMA")
+      SELECT a.attname
+        FROM (
+               SELECT conrelid, conkey, generate_subscripts(conkey, 1) idx
+                 FROM pg_constraint
+                WHERE conrelid = #{quote(quote_table_name(table_name))}::regclass
+                  AND contype = 'p'
+             ) c
+        JOIN pg_attribute a
+          ON a.attrelid = c.conrelid
+         AND a.attnum = c.conkey[c.idx]
+       ORDER BY c.idx
+    SQL
+  end
 end
 
 class ActiveRecord::ConnectionAdapters::PostgreSQLAdapter
--- a/ruby/rails/carrental/.gitignore
+++ b/ruby/rails/carrental/.gitignore
@@ -1,0 +1,2 @@
+/tmp/*
+!/tmp/.keep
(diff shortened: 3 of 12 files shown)


praba2210:
Reviewed at `40e3c63`. Fix looks correct, and the upsert tests catch a genuinely silent failure — pre-fix `updatable_columns` comes out empty and `configure_on_duplicate_update_logic` downgrades `:update` to `:skip`, so the upsert becomes `DO NOTHING` without erroring.

Six things, first three worth acting on:

| # | Where | Issue | Fix |
|---|---|---|---|
| 1 | [petclinic](https://github.com/aws-samples/aurora-dsql-samples/blob/40e3c63162b65fbcfe7fa5b16c4199887144bb94/ruby/rails/petclinic/Gemfile#L54) · [carrental](https://github.com/aws-samples/aurora-dsql-samples/blob/40e3c63162b65fbcfe7fa5b16c4199887144bb94/ruby/rails/carrental/Gemfile#L59) Gemfiles | `minitest "< 6"` is redundant — `activesupport` 7.2.3.1+ already declares `minitest >= 5.1, < 6`, `rails` pins it exactly, and both Gemfiles resolve into 7.2.3.x with no lockfile. Rails `main` dropped that bound for 8.1, so this will hold the samples on minitest 5 through a future upgrade | Drop both |
| 2 | [adapter.rb#L36](https://github.com/aws-samples/aurora-dsql-samples/blob/40e3c63162b65fbcfe7fa5b16c4199887144bb94/ruby/rails/petclinic/config/initializers/adapter.rb#L36) | README calls the override temporary, the code doesn't — and it's the only uncommented patch in that file | Comment citing rails/rails#58555; same link in the README |
| 3 | [owner.rb#L9-L13](https://github.com/aws-samples/aurora-dsql-samples/blob/40e3c63162b65fbcfe7fa5b16c4199887144bb94/ruby/rails/petclinic/app/models/owner.rb#L9-L13) + `pet`, `vet`, `specialty`, carrental `vehicle`, `customer`, `reservation` | "Without this, active record uses all the columns…" no longer describes what prevents it — `get_primary_key` → `schema_cache.primary_keys` → `SchemaStatements#primary_key` now unwraps `["id"]` itself | Keep the lines (fallback once the override goes), point them at the override |
| 4 | [carrental/dsql_adapter.rb#L7-L9](https://github.com/aws-samples/aurora-dsql-samples/blob/40e3c63162b65fbcfe7fa5b16c4199887144bb94/ruby/rails/carrental/config/initializers/dsql_adapter.rb#L7-L9) | "In development/test, SQLite3 is used instead" — but `test:` is `adapter: postgresql`, so the l.21 guard does activate there. Pre-existing | Fix the comment |
| 5 | [test_helper.rb#L8](https://github.com/aws-samples/aurora-dsql-samples/blob/40e3c63162b65fbcfe7fa5b16c4199887144bb94/ruby/rails/petclinic/test/test_helper.rb#L8) | `parallelize` + DSQL: `TestDatabases`' `after_fork_hook` → `reconstruct_from_schema` on `"postgres-#{i}"`, a drop/create DSQL lacks. Dormant — threshold is 50 tests, petclinic has 1 | Consider dropping; also matches carrental |
| 6 | commit c826e54 / PR body | `RAILS_ENV` switch is a CI no-op — `Runner.parse_options` sets `ENV["RAILS_ENV"] = "test"` unconditionally before boot, so only `db:migrate` was affected, and both configs are identical under `CLUSTER_USER: admin`. Body also omits the `fixtures :all` removal | Reword; add to the summary |

Suggested wording for 2:

```ruby
# Temporary until rails/rails#58555 ships: pg_index.indkey reports every column
# of DSQL's covering primary-key index, so read pg_constraint.conkey instead.
```

Also checked and fine: schema.rb matches the dumper's new output, dropping `fixtures :all` is safe (all five fixtures are untouched `MyString` scaffolds referenced by no test), and `ruby/rails/**` isn't in either sync workflow's rsync scope. `vet_specialty.rb` never had the `self.primary_key` line, and the phantom singular `owner` table predates this PR.

praba2210 (approved):

PR #1716, opened 2026-08-27
Foreign key feature release

davidrz15:
Part of the feature rollout with https://github.com/awslabs/aurora-dsql-tools/pull/142,
https://github.com/awslabs/aurora-dsql-orms/pull/598, and https://github.com/awslabs/agent-plugins/pull/261.


amaksimo:
Superseded by #1732. I rebased and rewrote this change so every sample defines foreign keys in its initial schema and removes the old application-level or upgrade paths. The replacement PR uses an upstream branch so the DSQL integration workflows can access repository credentials.

PR #1732, opened 2026-08-28
Use foreign keys in sample schemas

amaksimo:
Supersedes #1716 from an upstream branch so the integration workflows can access repository credentials. The commit retains the original author.

## Summary
- define foreign keys directly in each sample's initial schema or model
- remove application-level referential-integrity checks, upgrade migrations, and catalog validation
- delete README language that described foreign keys as unsupported
- keep normal database constraint error handling and enforcement tests

## Local verification
- auth/session: 196 tests and TypeScript build
- Drizzle: formatting, TypeScript build, and initial-migration test
- Prisma: formatting, client generation, and TypeScript build
- Sequelize: TypeScript build
- Go: internal package tests
- Tortoise ORM: 9 passed, 1 skipped
- SQLAlchemy: compiled metadata contains all three foreign keys
- Rails: 14 tests, 36 assertions
- DSQL lint: no diagnostics for the changed hand-written DDL
- repository audit: no remaining unsupported/application-level FK workaround language


anwesham-lab (commented):
### Code review

Repo: `aws-samples/aurora-dsql-samples` · PR: [#1732](https://github.com/aws-samples/aurora-dsql-samples/pull/1732) "Use foreign keys in sample schemas"
Head: `ddeebdf62f190ef8d20401106a1e2f615bb6f7c9` · Base: `84851bbbe14cddb3d34cd4f3731f4587d72d3a8d` · 45 files, +446 / −421
Review method: 23 scoped subagent lanes → consolidated → every finding re-validated individually. DSQL behavior verified by direct `psql` replay against a live cluster (us-east-1, PostgreSQL 16), **not** via `dsql-lint`/`dsql_recommend`, since that tooling is still being updated for FK support.

**Bottom line: 1 blocking issue, 1 strong recommendation, 3 minor notes, 1 follow-up. 12 candidate findings dropped as false positives.** The schema work itself is correct — every DDL artifact in this PR was replayed against the cluster and executes and enforces properly.

---


Verified DSQL's FK behavior directly against a live cluster rather than relying on `dsql-lint`, since that tooling is still catching up to FK support. Ground truth:

| Behavior | Result |
|---|---|
| Inline FK in `CREATE TABLE` (single-col, composite, self-ref, → UNIQUE) | Supported |
| `ALTER TABLE ... ADD CONSTRAINT ... FOREIGN KEY` | **Rejected** — `unsupported ALTER TABLE ADD CONSTRAINT statement` |
| `ON DELETE/UPDATE RESTRICT \| NO ACTION \| CASCADE \| SET NULL \| SET DEFAULT` | Accepted; CASCADE verified at runtime |
| Orphan INSERT / orphan UPDATE / DELETE of referenced parent | All raise SQLSTATE `23503` |
| `DEFERRABLE INITIALLY DEFERRED`, NULL FK values | Supported |
| `DROP TABLE` of a referenced parent | Blocked without CASCADE |
| Auto-index on child FK column | None created |

I replayed every DDL artifact in this PR against the cluster and it all executes and enforces correctly. One blocking issue.

---

#### Blocking

**1. `db/schema.rb` breaks `db:prepare` on a fresh DSQL cluster — a regression this PR introduces**

`add_foreign_key` sits outside `create_table`, so Rails renders it as `ALTER TABLE ... ADD CONSTRAINT ... FOREIGN KEY`, which DSQL rejects.

https://github.com/aws-samples/aurora-dsql-samples/blob/ddeebdf62f190ef8d20401106a1e2f615bb6f7c9/ruby/rails/carrental/db/schema.rb#L44-L48

The failure chain is the sample's own documented deploy path:

1. `bin/docker-entrypoint` runs `./bin/rails db:prepare` on every server start; `README.md:376` documents it ("On first boot ... Runs all database migrations (`rails db:prepare`)") and `README.md:437` tells local devs to run it too.
2. Rails 7.2 `DatabaseTasks.prepare_all` calls `load_schema` when `schema_migrations` doesn't exist yet — true for every fresh cluster. There's no `schema_format` override, so the default `:ruby` loads `db/schema.rb`.
3. `add_foreign_key` → `ALTER TABLE ADD CONSTRAINT` → rejected. `docker-entrypoint` is `#!/bin/bash -e`, so the task aborts before Puma starts and never recovers.

The migration itself is correct — `t.foreign_key` inside `create_table` renders inline, and I confirmed that path works on the cluster:

https://github.com/aws-samples/aurora-dsql-samples/blob/ddeebdf62f190ef8d20401106a1e2f615bb6f7c9/ruby/rails/carrental/db/migrate/20260327000003_create_reservations.rb#L13-L20

CI doesn't catch it because `.github/workflows/ruby-rails-carrental-integ-tests.yml:60` runs `bin/rails db:migrate`, never `db:prepare` or `db:schema:load`. Base `schema.rb` had zero `add_foreign_key` lines, so this path worked before this PR.

Suggested fix: drop the trailing `add_foreign_key` calls from the committed dump (keep FKs inline-only), or point the entrypoint at `db:migrate`. Note `schema_format = :sql` does **not** help — `pg_dump` also emits `ALTER TABLE ADD CONSTRAINT`.

---

#### Strongly recommended (upstream tooling lag, not this diff)

**2. `npm run db:generate` silently deletes the FKs this PR just added**

`typescript/drizzle/README.md:44-52` tells maintainers to run `npm run db:generate` after editing `src/schema.ts`. With the versions this PR pins (`@aws/aurora-dsql-drizzle@0.1.0`, `dsql-lint@0.2.14`) that command strips every FK:

```
WARNING — Table-level FOREIGN KEY constraint is not supported in DSQL.
  Removed table-level FOREIGN KEY constraint      (x3)
```

Root cause is upstream tooling that isn't FK-aware yet (support merged in `aurora-dsql-tools#147` but isn't in the pinned published version) — the committed `0000_init.sql` is correct and CI passes. But as it stands, the next person to follow the documented workflow loses all three constraints with only a stdout warning. Worth gating on the FK-aware `dsql-lint` release or adding a short note until then. `test/migrate.spec.ts` already guards the committed file against the `ALTER TABLE` form, which is a good touch.

---

#### Minor

**3. The new Go unit test never runs in CI.** `.github/workflows/go-recipe-api-integ-tests.yml:55` runs `go test -v ./test/...`, which cannot match `internal/store/errors_test.go`; there's no `go vet`/`go build ./...` either. `./...` fixes it. The new `IsForeignKeyViolation` → 409 handler path also has no coverage in `test/integration_test.go`.

**4. Prisma README dropped still-true limitations.** The removed bullet covered "no foreign keys, no sequences, no advisory locks" — only the first is obsolete. `package.json:16` still carries `PRISMA_SCHEMA_DISABLE_ADVISORY_LOCK=1` with nothing in the README explaining why. The migration-regeneration guidance and `@aws/aurora-dsql-prisma-tools` were removed wholesale, though the upstream fix (`aurora-dsql-orms#598`) was a 6-line change and the npm package simply isn't republished yet — worth saying so in the PR body. (Also: the PR body links `aurora-dsql-tools#142`, which is closed unmerged and superseded by `#147`.)

**5. Missing index on two FK columns the PR's own tests exercise.** `ratings.chef_id` (`internal/store/dsql.go:99-100` indexes `recipes.chef_id` and `ratings.recipe_id` but not `ratings.chef_id`, while `DeleteChef` is a live flow) and `reservations.customer_id` / `vehicle_id` (no index at all, and `test/controllers/foreign_key_delete_test.rb` deletes referenced parents). DSQL enforces RESTRICT correctly without them — it just full-scans the child table — so this is a perf nit, but it's inconsistent with the indexing pattern this PR applies elsewhere in the same files.

**6. Follow-up, not for this PR:** three other ports of this repo's canonical petclinic schema were left un-migrated — `typescript/type-orm` (0 FKs in `src/migrations/*.ts`, and `src/entity/Pet.ts:29` declares `onDelete: "CASCADE"` with no backing constraint, so it's a misleading no-op), `java/liquibase` (0 FKs in `changelog.sql`), and `ruby/rails/petclinic` (0 `add_foreign_key`, still `dependent: :destroy`/`:delete_all` — the opposite idiom from its sibling `ruby/rails/carrental` after this PR). Worth tracking issues before calling the initiative done.

---

#### Checked and cleared

Security (no ownership/tenancy check was removed — all were pure existence checks; no error-text leakage, no SQLi, no credential or workflow-permission changes) · every removed app-level check is backed by a real FK · statement ordering (parent before child) in all nine samples · one-DDL-per-transaction · no stale "FKs unsupported" language left in the touched samples or in `java/`, `rust/`, `dotnet/`, `php/`, `cpp/`, `javascript/`, `lambda/`, `cli/` · SQLAlchemy / Sequelize / Tortoise all emit inline FKs (no `ALTER TABLE` path, no FK cycles) · license headers, `frozen_string_literal`, NOTICE/LICENSE, `.gitignore` · Rails `belongs_to` semantics and controller `rescue ActiveRecord::InvalidForeignKey` · commit authorship claim holds.

`ruby/rails/carrental`'s new tests are the reference-quality example here — real cluster, both FK directions, model and controller layers.


---

## Validation record

### 2a. Validated findings

| # | Finding | Severity | How it was validated (incl. cluster checks) | Conf. |
|---|---|---|---|---|
| 1 | `ruby/rails/carrental/db/schema.rb:46-47` — `add_foreign_key` → `ALTER TABLE ADD CONSTRAINT`, rejected by DSQL; breaks `db:prepare` on a fresh cluster | **Blocking** | **Cluster:** replayed both Rails paths side by side. Path A (migration, `t.foreign_key` inline in `CREATE TABLE`) → 3× `CREATE TABLE` succeeded, then proved enforcement: orphan child INSERT → `23503`, DELETE of referenced parent → `23503`. Path B (the two `schema.rb` statements verbatim) → `ERROR: unsupported ALTER TABLE ADD CONSTRAINT statement`, twice. Also reproduced generically in an earlier probe. **Code:** fetched Rails `v7.2.2.1` `activerecord/lib/active_record/tasks/database_tasks.rb` and read `prepare_all` — `unless database_initialized` → `if File.exist?(schema_dump_path)` → `load_schema(db_config, ActiveRecord.schema_format, nil)`. Confirmed `grep -rn schema_format config/` → no override, so default `:ruby` → `db/schema.rb`. Confirmed `bin/docker-entrypoint:28` runs `./bin/rails db:prepare` under `#!/bin/bash -e`; README:376 and :437 document it. **Regression proof:** `git show 84851bb:.../db/schema.rb \| grep -c add_foreign_key` → `0`; diff adds exactly the 2 lines. **CI blind spot:** workflow line 60 runs `db:migrate` only. Converged independently in 4 lanes (Rails, README, comment, DDL-replay). | 98 |
| 2 | `typescript/drizzle` — documented `npm run db:generate` strips the new FKs | High (upstream) | **Executed** the PR's own pinned toolchain: `node node_modules/@aws/aurora-dsql-drizzle/dist/cli/index.js transform drizzle/0000_init.sql` → 3× `WARNING — Table-level FOREIGN KEY constraint is not supported in DSQL. Removed table-level FOREIGN KEY constraint`, and the emitted `pet` table has `"owner_id" UUID` with no constraint. Confirmed installed versions `aurora-dsql-drizzle@0.1.0`, `dsql-lint@0.2.14`; confirmed `package.json:8` `db:generate` = `aurora-dsql-drizzle generate` and README:44-52 instructs it. Cross-checked with the upstream history lane: FK support merged in `aurora-dsql-tools#147`, not in the pinned published build — so root cause is tooling lag, not this diff. Committed artifact verified correct by cluster replay. | 95 |
| 3 | `internal/store/errors_test.go` excluded from CI | Minor | `grep -rn "go test" .github/workflows/` → `go-recipe-api-integ-tests.yml:55: go test -v ./test/...`; `find go/recipe-sharing-api -name '*_test.go'` → `test/integration_test.go` + `internal/store/errors_test.go`. `./test/...` cannot match `internal/store`. No `go vet`/`go build ./...` in any workflow. Go toolchain absent locally, so verified by path semantics, not execution. | 90 |
| 4 | Prisma README deleted still-true DSQL limitations | Minor | `git diff` on README shows the removed bullet listing "no foreign keys, no sequences, no advisory locks"; `grep -i "advisory\|sequence"` on HEAD README → 0 hits; `package.json:16` still has `PRISMA_SCHEMA_DISABLE_ADVISORY_LOCK=1` (read directly). Upstream: `aurora-dsql-orms#598` diff = removal of one `checkRelationMode` function; `npm view @aws/aurora-dsql-prisma-tools versions` → only 0.1.0/0.1.2. PR-body link `aurora-dsql-tools#142` confirmed closed-unmerged. Two lanes agreed. | 85 |
| 5 | Missing index on `ratings.chef_id`, `reservations.customer_id`/`vehicle_id` | Minor (perf) | `grep -n "CREATE INDEX" internal/store/dsql.go` → only `idx_recipes_chef_id`, `idx_ratings_recipe_id`; `ratings.chef_id` FK exists without one while `DeleteChef` is live. `grep -n "add_index\|t.index"` across carrental `schema.rb` + migrations → none. **Cluster:** confirmed DSQL creates no auto-index on the child FK column (PK index only, with other cols in `INCLUDE`), so the RESTRICT check has no index to use. DSQL docs frame indexing as SHOULD (benchmark guidance), not MUST — hence minor, not blocking. | 85 |
| 6 | `type-orm` / `java/liquibase` / `rails/petclinic` un-migrated siblings | Follow-up | `grep -c "FOREIGN KEY\|REFERENCES" typescript/type-orm/src/migrations/*.ts` → `0`, while `src/entity/Pet.ts:29` has `onDelete: "CASCADE"` (no-op). `grep -ci "foreign key\|references" java/liquibase/changelog.sql` → `0`. `grep -c add_foreign_key ruby/rails/petclinic/db/schema.rb` → `0`; `petclinic/app/models/{owner,vet}.rb` still use `dependent: :destroy` / `:delete_all`. `git diff --name-only` confirms none are touched by this PR. | 90 |

### 2b. Dropped findings — false positives

| Claimed finding | Why dropped |
|---|---|
| `.github/workflows/typescript-prisma-integ-tests.yml` **deleted** (57 lines, 0 added) | **My own initial framing was wrong.** `git cat-file -e HEAD:<path>` → file exists; `git diff --summary` lists no delete mode. The 57 lines are 3 removed CI steps inside an otherwise-unchanged file. The `create-cluster → test → delete-cluster` lifecycle is intact and `gh pr checks 1732` shows `prisma-integ-test` passing. |
| Drizzle `createPet` lost OCC-retry protection (2 lanes called this blocking) | Read before/after: the `transactionWithRetry` wrapper existed only to hold the removed read-then-write. `createOwner` and `createSpecialty` are bare single-statement inserts with no retry, unchanged by this PR — so `createPet` now *matches* the sample's established convention rather than deviating. Not a PR-introduced inconsistency. Deletion-audit lane concurred. |
| `sessionRepository.create` FK violation → opaque 500 instead of 401 (2 lanes) | Traced reachability: `sessionService.createSession` is only ever called with a `userId` fetched and authenticated in the same request, and no delete-user path exists anywhere in the sample (`grep` for `deleteUser`/`DELETE FROM users` → nothing). Unreachable via the public API; only a theoretical TOCTOU. |
| Deno `booking-api` lost referential integrity (0 lines added, 3 removed) | The 3 removed lines are a comment. The sample has exactly one table (`bookings`); `booked_by` is free-text used as a test-run tag, with no `users` table to reference — before or after. No check existed to lose. |
| SQLAlchemy may emit `ALTER TABLE ADD CONSTRAINT` for circular FKs | Compiled `CreateTable` against the postgresql dialect in a real venv: all 3 FKs inline, `sorted_tables` = `owner, specialty, vet, pet, specialty_to_vet` (correct topological order), no cycles, no `use_alter=True`. |
| Sequelize may emit `ALTER TABLE ADD CONSTRAINT` via `sync({alter:true})` | `sync()` is never called; schema is built with explicit `queryInterface.createTable`. Drove Sequelize's actual Postgres QueryGenerator and captured inline `REFERENCES ... ON DELETE RESTRICT ON UPDATE RESTRICT`. Creation order is dependency-correct. |
| Prisma `relationMode` change incomplete / inconsistent | `relationMode = "prisma"` removed (defaults to `foreignKeys`) and `OrderItem.order` gained `onDelete: Restrict, onUpdate: Restrict`, exactly matching the inline SQL. Consistent. |
| Removed `validate`/`dsql-migrate` scripts leave dangling references | Repo-wide grep for `npm run validate`, `dsql-migrate`, `aurora-dsql-prisma`, `constraints: false`, `restrict_with_error`, `primaryjoin` → no surviving references. Clean removal. (The *documentation* loss is captured as validated finding #4.) |
| A removed existence check was really an authorization/tenancy check (IDOR) | Every removed check inspected across Go, Rails, Drizzle and session-mgmt: all pure `WHERE id = ?` existence checks, no `owner_id`/session comparison. The one real ownership check (`revokeByIdForUser`, `WHERE id = $1 AND user_id = $2`) is untouched. |
| Handlers leak DSQL `23503` DETAIL (table/column/key values) to clients | Every changed handler returns canned strings ("chef no longer exists", "Cannot delete customer with reservations."). No `err.Error()`/`e.message` reaches a response body. Session-mgmt falls through to a generic 500 with no message included. |
| Nullable FK paired with `ON DELETE SET NULL`, or ORM/DB nullability divergence | `grep -in "set null\|ondelete\|on_delete\|cascade"` over the whole diff: every FK added uses `RESTRICT` on both actions; no `SET NULL` anywhere. Nullability matches between ORM and DDL on every touched FK column. |
| Missing Apache-2.0 headers / NOTICE / CHANGELOG / PR-template steps | Sampled sibling files to establish actual convention: Go files carry the repo's `MIT-0` header and do; Ruby files carry `frozen_string_literal: true` and do; TS siblings have no header by convention. No CHANGELOG exists in the repo. Only dependency change is a removal, so NOTICE is unaffected. |
| Repo still contains "FKs unsupported" prose | Repo-wide grep across all languages incl. untouched `java/`, `rust/`, `dotnet/`, `php/`, `cpp/`, `javascript/`, `lambda/`, `cli/` → clean. Only near-miss is `sample-amazon-aurora-dsql-agent/agent/database_agent.py:66` ("no formal FKs in DSQL"), which is out of this PR's scope and arguably describes that schema; folded into follow-up #6 rather than reported as a defect. |

### 2c. Cluster operations performed

All against a scratch DSQL cluster (us-east-1), admin token via `aws dsql generate-db-connect-admin-auth-token`, `sslmode=verify-full` with `sslrootcert=system`.

- **Capability probe:** inline FK (single-col / composite / self-ref / → UNIQUE), `ALTER TABLE ADD CONSTRAINT` (rejected), all five referential actions, `DEFERRABLE INITIALLY DEFERRED`, `CREATE INDEX ASYNC`, index inventory (confirmed no auto-index on child FK column).
- **Enforcement probe:** orphan INSERT → `23503`; valid INSERT → OK; DELETE of referenced parent → `23503`; NULL FK → OK; UPDATE to orphan → `23503`; CASCADE delete → child removed; deferred constraint satisfied at COMMIT; `DROP TABLE` of referenced parent → blocked; `TRUNCATE` → unsupported (pre-existing, unrelated).
- **Rails A/B replay:** inline (`db:migrate`) path succeeded and enforced both FK directions; `schema.rb` `ALTER TABLE` path rejected twice.
- **Subagent replays:** all nine samples' DDL artifacts executed verbatim or in exact-emitted shape; also confirmed empirically that two DDL statements in one transaction fail (`multiple ddl statements not supported in a transaction`).
- **Cleanup:** all probe schemas dropped (one `DROP SCHEMA` hit `OC001` — DSQL's OCC in action — and succeeded on retry). Final schema list is `fkval, ix, mf, public, sys, v2`, all pre-existing and untouched.

### 2d. Note on tooling accuracy

The DSQL lint/plugin tooling is still being updated for FK support, so no finding here rests on `dsql_lint` / `dsql_recommend` / `dsql_search_documentation` output. Every DSQL behavioral claim comes from `psql` execution against the live cluster. That caution proved well-founded: finding #2 exists precisely *because* the shipped `dsql-lint@0.2.14` still reports FKs as unsupported and strips them.

---

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anwesham-lab:
Of the above findings, the ruby on rails finding and resolving DSQL Lint to the upgraded 0.2.17 since that's the first documented FK support are the two changes I think block my approval. All others can be done in follow-up for cleanliness.

PR #1845, opened 2026-09-14
feat(javascript): Add Sequelize ORM sample for Aurora DSQL

lasitab-code:

## Summary
  Adds a Node.js sample demonstrating Sequelize ORM with Aurora DSQL, including:
  - CRUD operations for a hotel reservation domain (guests, rooms, reservations, payments)
  - IAM authentication (admin + non-admin user flows) via the Aurora DSQL node-postgres connector (`dialectModule`), with automatic token generation
  and Region discovery
  - Foreign keys between the hotel entities
  - OCC retry with exponential backoff and jitter
  - Schema isolation for a least-privilege, non-admin application user
  - Unit tests (Jest) and live integration tests

  ## Testing
  - End-to-end tested with admin and non-admin users against a live DSQL cluster (us-west-2)
  - Unit tests pass: `npm test` (18 passed)
  - Integration tests pass against a live cluster: `npm run test:integration` (5 passed); skip cleanly without credentials
  - Security scan: 0 CVEs, 0 secrets, 0 code vulnerabilities


javascript/sequelize/src/app.js
```diff
@@ -0,0 +1,274 @@
...
+  // Cleanup (commented out - use cleanup.js separately)
+  // console.log('\n▶ Cleaning up...');
+  // await cleanup(Guest, Room, Reservation, Payment);
+
+  // Close connection
+  await sequelize.close();
+  console.log('\n✓ Demo completed successfully!');
+}
+
+main().catch((err) => {
```
praba2210, line 271: No test executes this file — `main()` runs at import and nothing is exported, so both suites only test the helpers. Verified: injected a ~24% billing bug and a crash bug; all 23 tests stayed green both times.

Suggest the sibling pattern (node-postgres, batch_operations):
```js
export { main };
if (import.meta.url === `file://${process.argv[1]}`) {
  main().catch((err) => { console.error(err); process.exit(1); });
}
```
then run `main()` from the integration test.

javascript/sequelize/src/db.js
```diff
@@ -0,0 +1,155 @@
...
+    dialect: 'postgres',
+    dialectModule: {
+      ...pg,
+      Client: AuroraDSQLClient,
+    },
+    dialectOptions: {
+      clientMinMessages: 'ignore',
+    },
+    // Route all queries to the selected schema (public for admin, custom for non-admin).
+    searchPath: SCHEMA,
```
praba2210, line 62: `searchPath` makes Sequelize disable bind parameters — every model INSERT/UPDATE ships values inlined as literals (verified on the wire). Deleting this line restores `$1, $2, …` and queries stay schema-qualified via each model's `schema:` option. Ran the sample without it — identical behavior.

javascript/sequelize/README.md
```diff
@@ -0,0 +1,235 @@
...
+
+### 2. Create the database role and schema (run once as admin)
+
+Edit `src/setup_app_user.sql` — replace `<AWS_ACCOUNT_ID>` and `<IAM_USERNAME>` with your values, then run:
+
+```bash
+# Generate an admin auth token
+TOKEN=$(aws dsql generate-db-connect-admin-auth-token \
+  --hostname $CLUSTER_ENDPOINT \
+  --region $CLUSTER_REGION \
```
praba2210, line 115: Three documented steps fail as written:
1. Here — `CLUSTER_REGION` is "optional" per line 45 but this snippet requires it.
2. Line 213 — FK query filters `table_schema = 'public'`; the recommended `hotel`-schema path returns 0 rows vs "Expect four foreign keys".
3. `setup_app_user.sql:38` — "run the app with CLUSTER_USER=admin" can't create the `hotel` tables; `db.js:28-30` forces admin to `public`.

praba2210 (changes requested):
Ran end-to-end on a real DSQL cluster - clean run, 4 FKs enforced, real OCC retry. Substance is solid; blockers are inline. Non-blocking:

| Type | Where | Issue | Fix |
|---|---|---|---|
| should-fix | `README.md:180` | "write paths use the OCC retry helper" — only 1 of ~10 writes does. Wrap or soften. | Wrap the multi-step writes in `withOccRetry`, or reword to "…surface as retryable serialization errors — see `retry.js`, demonstrated on the loyalty-tier update". |
| should-fix | `package.json` | `@aws-sdk/credential-providers` + `dsql-signer` never imported — drop; regen lockfile (`npm ci` errors today). | `npm uninstall @aws-sdk/credential-providers @aws-sdk/dsql-signer` — drops both and syncs the lockfile in one step (verified: `npm ci` then passes, unit tests green; the connector still pulls both transitively). |
| should-fix | `setup_app_user.sql:46-49` | ADP rationale wrong: owners can't be denied own tables, and admin's ADP never covers app-created ones. Keep grant, fix comment. | Replace "Without this, the app hits permission denied on its own tables." with: this covers tables **admin** creates later in the schema; tables the app creates it already owns. |
| nit | `db.js:78` | `createSchema()` creates tables → `createTables()`. | 7 occurrences in 3 files: `db.js:78`,`155`; `app.js:6`,`189`; `test/integration.test.js:32`,`39`,`45`. |
| nit | `README.md:183` | Connector's occ-retry only wraps `connect()`; `retry.js` is the real mechanism — worth a line. | Add a clause to adaptation 5 — it explains why `retry.js` exists. |
| nit | `test/test_example.test.js:39` | Backoff timing unasserted (counts/give-up well covered). | Spy on `setTimeout`, force 3 OCC failures, assert each delay is in `[50·2^i, 100·2^i)` — covers exponent and jitter. |
| nit | headers | SPDX `Apache-2.0` vs package.json `MIT-0` — align. | Switch the 8 headers to `MIT-0` (matches `LICENSE`, `package.json:36`, `README.md:235`), though other samples are mixed. |

javascript/sequelize/src/app.js
lasitab-code: Fixed in d31d2d4. app.js now exports main() and only auto-runs under an `import.meta.url === \file://${process.argv[1]}`guard, so importing it has no side effects.integration.test.jsimportsmain and runs it in a new final test (main() runs the full demo end to end`), giving the orchestration real coverage. 
Verified live against a DSQL cluster: 6/6 integration tests pass, including the new one (2.4s).

javascript/sequelize/src/db.js
lasitab-code: Removed the searchPath: SCHEMA option; schema targeting stays via each model's schema: in models.js, so queries remainschema-qualified while bind parameters are restored. Verified end to end on a live cluster on both the admin and non-admin (hotel_app/hotel) paths — npm start and npm run cleanup both succeed, and the schema-qualified .destroy() cleanup works for the non-admin user.

javascript/sequelize/README.md
lasitab-code: 1. Added an explicit export CLUSTER_REGION=... before the admin token snippet, with a note that the AWS CLI token command needs --region even though the app auto-detects it.

2. The FK verification query no longer filters table_schema = 'public' — it now selects table_schema and orders by it, so it works for both the public (admin) and hotel (non-admin) paths; the prose notes the schema each FK appears under.
 
3. Corrected setup_app_user.sql step 4: for strict least-privilege, admin creates the hotel tables directly via SQL/psql (running the app as admin can't, since db.js forces admin to public).

lasitab-code:
Thanks for the thorough review. All items addressed in d31d2d4:

- should-fix OCC-retry wording: reworded to scope it to the loyalty-tier update and point at retry.js.
- should-fix unused deps: dropped @aws-sdk/credential-providers and @aws-sdk/dsql-signer and regenerated the lockfile — npm ci now passes (it errored before).
- should-fix ADP rationale comment corrected.
- nit createSchema() → createTables() (all refs + adjacent comment/log).
- nit added a note that the connector's built-in OCC retry only applies to its transaction() method (not the Sequelize dialectModule path), which is why retry.js exists.
- nit added a backoff-timing test asserting each delay ∈ [50·2ⁱ, 100·2ⁱ).
- nit aligned all 8 SPDX headers to MIT-0.
Verification: npm ci clean, npm test 19/19, npm run test:integration 6/6 on a live cluster (admin + non-admin paths).

praba2210 (approved):
Looks good!

PR #1854, opened 2026-09-15
fix(recipe-sharing-api): use reserved amzn-s3-demo- bucket prefix

arbeiterz:
Rename the S3 deployment bucket to use the reserved 'amzn-s3-demo-' prefix so it follows the AWS documentation naming convention for example buckets.

The bucket name is derived from STACK_NAME in deploy.sh, so prepending the prefix to the derived name preserves the --stack-name override and propagates the new name to bucket creation, package upload, the LambdaS3Bucket CloudFormation parameter, and the tear-down hint. The matching cleanup command in README.md is updated to the same name.


amaksimo:
This is already in `main` via #1675. Also, S3 rejects the `amzn-s3-demo-` prefix for general-purpose buckets. We should close this and fix `main`.

amaksimo:
Replaced by #1857, which fixes this on current `main`.

PR #1855, opened 2026-09-15
docs(recipe-sharing-api): fix README and deploy accuracy

arbeiterz:
- README: add the test/ directory (integration_test.go) to the Project Structure so it reflects the actual code layout.
- deploy.sh: require Go 1.26+ to match the 'go 1.26.3' directive in go.mod (the check previously allowed 1.24, which cannot build the module).
- cloudformation.yml: remove the unused DB_TYPE Lambda environment variable; no Go code reads it (both entrypoints read only DSQL_ENDPOINT).


amaksimo:
`go.mod` requires 1.26.3, but this accepts 1.26.0–1.26.2 when `GOTOOLCHAIN=local`. Can we drop this check and let Go enforce `go.mod`?

arbeiterz:
Good catch, thanks — dropped the check in 31ef443.

You're right on both counts. The sed only captured major.minor (go1.26.3 → 1.26), so the comparison never looked at the patch component at all. It would happily pass 1.26.0–1.26.2 despite go.mod requiring 1.26.3, and as you noted, under GOTOOLCHAIN=local that's exactly the case where nothing else would catch it either.

Rather than fix the parser to compare the full major.minor.patch against a hardcoded 1.26.3, I removed the gate entirely. The Go toolchain already enforces the go directive in go.mod — including the patch version, in every GOTOOLCHAIN mode — so the manual check was both redundant and a second source of truth that had already drifted out of sync with go.mod. Letting Go own it keeps go.mod as the single place the minimum version is declared, and the go build / go mod tidy calls later in the script surface a clear error if the toolchain is too old.

I kept the GO_VERSION capture (now preserving the patch) purely for the "Prerequisites verified" info line, and command -v go still guards that Go is installed.

go/recipe-sharing-api/deploy.sh
```diff
@@ -142,14 +142,10 @@ if ! command -v jq &>/dev/null; then
...
-GO_MAJOR=$(echo "$GO_VERSION" | cut -d. -f1)
-GO_MINOR=$(echo "$GO_VERSION" | cut -d. -f2)
-if [[ "$GO_MAJOR" -lt 1 ]] || { [[ "$GO_MAJOR" -eq 1 ]] && [[ "$GO_MINOR" -lt 24 ]]; }; then
-  err "Go 1.24 or later is required (found go${GO_VERSION})"
-  exit 1
-fi
+# The minimum Go version is declared by the `go` directive in go.mod and is
+# enforced by the Go toolchain itself (including the patch version, in every
+# GOTOOLCHAIN mode), so we don't re-check it here.
+GO_VERSION=$(go version | sed 's/.*go\([0-9][0-9.]*\).*/\1/')
```
amaksimo, line 148: Can we validate `go.mod` before printing "Prerequisites verified"? With `GOTOOLCHAIN=local` and Go 1.25.5, this prints success, then reports the version failure as a proxy failure.

amaksimo (changes requested):
The unresolved Go prerequisite issue still reproduces at this revision: with `GOTOOLCHAIN=local` and Go 1.25.5, the script reports prerequisites verified, then treats the `go.mod` version failure as a proxy failure. Please validate the module requirement before printing the success message; the existing inline comment has the details.

go/recipe-sharing-api/deploy.sh
arbeiterz: Good catch — fixed in 4cc1736.

The root cause was that "Prerequisites verified" printed before anything actually checked the running Go against go.mod, and the later `go mod tidy` had its stderr suppressed (2>/dev/null), so a version mismatch fell through to the GOPROXY=direct fallback and surfaced as a proxy error rather than a version error.

I added a validation step before the success message: it runs `go list -m` (lightweight, no network) which triggers the toolchain's version-directive enforcement, and on a version failure reports "This module requires Go 1.26.3 or later, but the active toolchain is goX" with a hint to upgrade or unset GOTOOLCHAIN. I kept it so Go does the version comparison rather than re-parsing versions in the script. So with GOTOOLCHAIN=local and Go 1.25.5 it now fails fast with a clear message.

